PR #47, opened 2026-05-22
fix: cursor-plugin @latest tag, http retry, and login_url scheme validation

ashish993:
## Changes

**Bug fix – release workflow**
- Add `Initialize Cursor latest ref` and `Update Cursor latest ref` steps to `.github/workflows/release.yml`. The `@call-e/cursor-plugin@latest` git tag was never initialized or updated on publish, unlike the Codex and Claude plugin equivalents.

**Security – broker login_url validation** (`packages/core/lib/broker-client.js`)
- Validate that the `login_url` returned by the broker server uses `https:` before it is passed to `openBrowser`/`spawn`. Loopback addresses are exempt to preserve local-dev and test compatibility.

**Reliability – HTTP retry with exponential backoff** (`packages/core/lib/http.js`)
- Retry transient HTTP errors (429, 500, 502, 503, 504) and network-level failures up to 2 times with 500 ms / 1000 ms delays before re-throwing.

## Testing
All 40 existing unit and e2e tests pass (`pnpm test && pnpm check`).

### DIFF
--- a/.changeset/fix-http-retry-and-login-url-validation.md
+++ b/.changeset/fix-http-retry-and-login-url-validation.md
@@ -1,0 +1,5 @@
+---
+"@call-e/core": patch
+---
+
+Add exponential backoff retry for transient HTTP errors (429, 5xx) and network failures in `requestJson`. Validate that `login_url` returned by the broker server uses the `https:` scheme before opening it in a browser (loopback addresses are exempt to support local development).
--- a/.changeset/fix-path-redaction-and-windows-opener.md
+++ b/.changeset/fix-path-redaction-and-windows-opener.md
@@ -1,0 +1,5 @@
+---
+"@call-e/cli": patch
+---
+
+Fix Windows shell injection in browser opener (use rundll32 instead of cmd /c start), remove home-directory paths from public JSON outputs (cache_path, pending_cache_path), omit --cache-root from suggested commands when using the default location, and restore safe URL sanitization in all login-URL output fields.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,17 +12,17 @@ jobs:
 
     steps:
       - name: Check out repository
-        uses: actions/checkout@v5
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
 
       - name: Set up pnpm
-        uses: pnpm/action-setup@v5
+        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5
         with:
           version: 10.18.3
 
       - name: Set up Node.js
-        uses: actions/setup-node@v5
+        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5
         with:
-          node-version: 22
+          node-version: 24
           cache: pnpm
 
       - name: Install dependencies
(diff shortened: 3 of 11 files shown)


Ray-56 (commented):
Thanks for the PR. I would hold off on merging this until the retry behavior is tightened up. I found a few issues:

1. `requestJson` now retries every non-`HttpStatusError` thrown after `fetch`, including local response parsing/validation failures. For example, a 200 response with malformed JSON is retried twice before throwing; I reproduced this with an invalid JSON response and saw 3 total requests. Those are not transient network failures, so malformed/invalid responses should fail fast.

2. Retries apply to every HTTP method. This helper is used by broker `POST /sessions` and `POST /exchange`, so a transient 5xx or network failure can duplicate non-idempotent operations. I would restrict retries to idempotent methods by default, or make POST retries an explicit opt-in where the caller knows it is safe.

3. The IPv6 loopback exemption does not currently work as described. In Node, `new URL("http://[::1]/").hostname` returns `[::1]`, not `::1`, so `http://[::1]/...` is rejected even though the PR says loopback addresses are exempt.

Validation I ran locally on a temporary worktree:
- `pnpm test`
- `pnpm run check:versions`
- `pnpm -r --filter "./packages/*" run check`

Those passed. I also checked the PR branch with `gh pr checks 47`, and GitHub currently reports no checks for this PR branch.

ashish993 (commented):
Review this. @Ray-56

Ray-56 (changes requested):
The latest commit addresses the previously reported retry issues: malformed JSON now fails fast, unsafe POST requests are not retried, and bracketed IPv6 loopback handling is covered. Two current-branch issues still prevent merge.

[P1] Rebase onto the current `main` and resolve the merge conflict without rolling package metadata or declarations backward. GitHub currently reports this PR as conflicting, and the branch still carries older Core package state than current main. After rebasing, rerun Core check/test/typecheck, `pnpm run check:versions`, and the pack dry-run so the reviewed result matches what would actually merge.

[P2] Split or remove the Cursor `@latest` release-workflow change unless maintainers explicitly decide to introduce that public alias. This repository currently documents automatic `@latest` maintenance for Codex and Claude only, and no Cursor install path consumes `@call-e/cursor-plugin@latest`. Combining a token-bearing release workflow change with Core retry/security behavior also makes this PR unnecessarily broad. Prefer a separate PR that updates `docs/agent-integration-layout.md`, explains the supported Cursor install path, and validates the release behavior; otherwise keep this PR limited to Core.

The existing `@call-e/core` patch changeset is appropriate for the runtime portion.

Ray-56 (changes requested):
Thanks for the update. The current head is still blocked.

[P1] The Windows browser opener still spawns the unqualified executable name `rundll32.exe`. On Windows this can resolve through the current working directory before System32, so a planted executable in an untrusted checkout can run during `calle auth login`. Use the fully qualified `%SystemRoot%\System32\rundll32.exe` path and test the production command-selection path, or remove this overlapping change and rely on PR #71.

[P1] The branch no longer contains the Cursor `@latest` initialization/update behavior described by its title and body. Instead it now mixes CLI path redaction and browser opening, cache migration, Core retry/type changes, CI runtime changes, and action pinning. Please rebuild this branch from current `main` and keep it to one objective. If the Cursor alias is still intended, use a dedicated PR with the required integration-layout documentation and release-policy decision.

[P2] The newly added cache migration is currently MD5-to-MD5: both `serverHash` and `legacyServerHash` use MD5, so the migration is a no-op. Remove it unless this PR intentionally changes the current hash and supplies complete migration coverage.

GitHub CI did not execute for this head: the run ended as `action_required` with zero jobs. Please rerun CI after the branch is narrowed. The Core and CLI patch changesets are present, but their final scope should match the rewritten PR.

PR #49, opened 2026-05-24
security: enterprise review fixes — HTTPS enforcement, input limits, path redaction, CI alignment

ashish993:
## Summary

Enterprise security review fixes across `packages/cli`, `packages/core`, and `.github/workflows`. All changes pass `pnpm check` and `pnpm test` (27 unit + 13 E2E tests).

### Critical
- **C-2**: Remove `cache_path` / `pending_cache_path` (full home-directory paths) from all public JSON outputs (`publicPendingLoginPayload`, `publicLoginPayload`, `statusPayload`). `statusPayload` now also validates `pending_login_url` is HTTPS before including it.
- **C-3**: Validate `login_url` from broker is HTTPS before writing to stderr or passing to `openBrowser`.

### High
- **H-1**: Add 64 KB max-size guard on `--args-json` in `parseJsonObject()` to prevent oversized payloads.
- **H-2**: Cap `--to-phone` at 10 numbers and `--goal` at 2000 characters in `buildPlanArguments()`.
- **H-4**: Enforce HTTPS on all user-supplied base URLs (`--base-url`, `--server-url`, `--broker-base-url`, `--auth-base-url`). Loopback addresses (`localhost`, `127.0.0.1`) are allowed with HTTP for local development/testing.

### Medium
- **M-5**: Align CI workflow to Node 24 (was 22) to match the release workflow.

### Low
- **L-1**: Truncate server-controlled error messages to 200 chars and strip newlines in `McpHttpError` to prevent server-injected content from reaching LLM context.
- **L-2**: `CALLE_TELEMETRY_URL` misconfiguration no longer throws and blocks all CLI commands — warns to stderr and disables telemetry gracefully.
- **L-4**: `loginCommand()` omits `--cache-root` from the hint string when it equals the default (`~/.calle-mcp/cli`), avoiding home-directory path leakage in assistant hints.

## Testing

```
pnpm check   # all syntax/lint checks pass
pnpm test    # 27 unit + 13 E2E tests pass
```


Ray-56 (changes requested):
Reviewed locally. I recommend holding this PR until the following issues are addressed.

1. `requestJsonRpc` retries non-idempotent `tools/call` requests.

`packages/core/lib/mcp-client.js` now retries every JSON-RPC request on 429/502/503/504 and some thrown fetch errors. That includes `tools/call`, which can execute non-idempotent operations such as `run_call`. If the server processes the first request but returns a transient 5xx, or if the response is lost, the retry can trigger duplicate side effects such as duplicate phone calls.

I reproduced this locally with a fake MCP server: the first `tools/call` returned 503, the client retried, and the server saw `toolCalls: 2`. Please restrict retries to safe/idempotent MCP methods such as `initialize` / `tools/list`, or add an explicit idempotency mechanism before retrying tool calls.

2. Broker-controlled `login_url` is still returned verbatim in public JSON outputs.

The PR sanitizes the URL used in `assistant_hint`, stderr, and browser opening, but `publicPendingLoginPayload()` and `authRequiredPayload()` can still include the raw `pending.login_url` / cached pending `login_url` as top-level JSON fields. I reproduced `auth login --start-only` returning:

```json
"login_url": "file:///tmp/injected"
```

That reintroduces the unsafe server-controlled URL into the public surface this PR is trying to harden. Please reuse the same HTTPS-only sanitizer for any public `login_url` field, or omit the field when the URL is not safe.

Validation run on the PR head in a detached worktree with Node v22.21.1:

- `pnpm test` passed
- `pnpm check` passed
- `pnpm pack:dry-run` passed

Release note: because this changes published behavior in `@call-e/cli` and `@call-e/core`, a patch changeset for both packages is recommended before merge.

Ray-56 (changes requested):
I re-reviewed the latest head against current main. The earlier non-idempotent MCP retry and raw public login_url findings are addressed, and the pinned Action SHAs resolve to legitimate upstream releases. The following issues still block merge:

[P1] Windows command injection remains in the browser-opening path. After validating only the URL scheme, the code still calls `cmd /c start "" <url>`. A valid HTTPS OAuth URL can contain `&`, which cmd.exe treats as a command separator; a compromised broker can therefore turn a validated URL into local command execution. Replace this with a shell-free opener using a trusted absolute system executable/API and cover the production path. Coordinate this with #71 rather than keeping two competing fixes.

[P1] Do not change the cache-directory hash without a migration and cleanup path. Switching MD5 to SHA-256 changes every `tokenCachePath` and `pendingCachePath`, silently logs Core consumers out, strands the previous token on disk, and makes the new logout path unable to delete the legacy credential. MD5 here is only a deterministic directory name, not an authenticity primitive, so the change adds no meaningful security. Either keep the current hash or implement secure read/migrate/remove support for both paths with tests.

[P1] Rebase onto current main and update the public TypeScript declarations. Current main now ships `.d.ts` files, but this PR exports `isSafeBrokerLoginUrl` and `sanitizeBrokerLoginUrl` without adding them to `broker-client.d.ts`. The runtime and published type surface would disagree.

[P2] Use one login-URL policy end to end. `sanitizeBrokerLoginUrl` accepts HTTP loopback URLs, while the default CLI `openBrowser` rejects every non-HTTPS URL. A locally supported URL can therefore pass broker validation and then fail at the actual browser-open step. Share the same predicate and add an end-to-end loopback test.

[P2] Complete the path-redaction work. `callStatusCommand` still embeds `config.cacheRoot` in public `next_command` output, and `auth logout` still returns `cache_path` and `pending_cache_path`. These continue to expose the user's home path through agent-visible JSON.

[P2] Add patch changesets for `@call-e/cli` and `@call-e/core`, then rerun the full check/test/pack suite after the rebase.

ashish993 (commented):
@Ray-56  — All findings from your latest review have been addressed in the latest push (d59a447):

[P1] Windows injection: [openBrowser](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) now uses [isSafeBrokerLoginUrl](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) guard + rundll32.exe (shell-free) on Windows
[P1] MD5→SHA-256 migration: legacyServerHash + [migrateTokenCache](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) added to cache.js with 3 tests
[P1] Type declarations: [isSafeBrokerLoginUrl](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [sanitizeBrokerLoginUrl](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) added to [broker-client.d.ts](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
[P2] URL policy: [openBrowser](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) now uses same [isSafeBrokerLoginUrl](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) predicate as broker validator (HTTP loopback accepted)
[P2] Path redaction: auth logout, [callStatusCommand](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [loginCommand](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) no longer expose [cacheRoot](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
[P2] Changesets: [@call-e/cli](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) patch changeset added; [@call-e/core](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) already had one
All 7 package test suites pass ([pnpm check && pnpm test](vscode-file://vscode-app/Users/ashishsharma/Desktop/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)). Ready for re-review.

Ray-56 (changes requested):
Thanks for addressing several earlier findings. The current head is still blocked.

[P1] Rebase onto current `main` and resolve the add/add declaration conflict without rolling package metadata backward. This branch still carries `@call-e/core` 0.2.2 and drops the current `types` entry and `check:types` script from 0.2.3. Preserve the declarations and type-checking state already on `main`, then rerun all package checks and pack validation.

[P1] The Windows browser opener still spawns the unqualified executable name `rundll32.exe`. That retains the current-directory binary-planting risk. Use the fully qualified System32 path and test the production command-selection path, or remove this overlapping change after PR #71 lands.

[P1] The MD5-to-SHA-256 cache migration only runs inside `loginWithBroker`. Existing users invoking `auth status`, `auth login --start-only`, MCP/call commands, or logout read only the new SHA-256 path and will ignore their existing MD5 token or pending session. Either retain the existing cache key or centralize a safe migration before every cache access, with upgrade tests for each public entry point.

[P2] Add a patch changeset for `@call-e/core`. The CLI patch changeset is present, but the Core retry, cache, broker, and MCP behavior is also user-visible.

There is no CI result for the current head. Please rerun CI after the rebase and conflict resolution.

PR #77, opened 2026-08-04
fix(deps): enforce a seven-day pnpm release age

anupamme:
## Summary

- Adds `minimumReleaseAge: 10080` to `pnpm-workspace.yaml`, requiring newly published package versions to be at least seven days old before pnpm will install them
- Hardens dependency resolution for this repository's own development and release installs; downstream consumers of published `@call-e/*` packages control their own package-manager policy independently

## Vulnerability

| Field | Value |
|-------|-------|
| **ID** | package_managers.pnpm.pnpm-missing-minimum-release-age.pnpm-minimum-release-age |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `package_managers.pnpm.pnpm-missing-minimum-release-age.pnpm-minimum-release-age` |
| **File** | `pnpm-workspace.yaml:1` |
| **Assessment** | Defensive hardening |

**Description**: Without a minimum release age, a newly published (potentially malicious) package version could be installed immediately. Setting `minimumReleaseAge: 10080` (7 days in minutes) adds a time-based buffer. Supported since pnpm v10.16.0.

## Changes

- `pnpm-workspace.yaml` — adds `minimumReleaseAge: 10080`

## Behavior Preservation

Only affects which package versions pnpm will resolve during install. Valid, established package versions (older than seven days) are unaffected.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,5 +2,3 @@ packages:
   - packages/*
 
 minimumReleaseAge: 10080
-blockExoticSubdeps: true
-trustPolicy: no-downgrade
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,2 +1,6 @@
 packages:
   - packages/*
+
+minimumReleaseAge: 10080
+blockExoticSubdeps: true
+trustPolicy: no-downgrade


Ray-56 (changes requested):
Thanks for the focused supply-chain hardening. The current head is blocked by a toolchain compatibility gap.

[P2] Two of the three settings do not take effect with the pnpm version this repository pins. package.json specifies pnpm@10.18.3, while pnpm documents trustPolicy as added in 10.21.0 and blockExoticSubdeps as added in 10.26.0: https://pnpm.io/settings/dependency-resolution

The current pnpm 10.18.3 install and repository checks pass without enforcing those two controls, which creates a misleading sense that all three protections are active. minimumReleaseAge is supported since 10.16.0 and is the only effective setting in this branch.

Please either:
1. Keep this PR limited to minimumReleaseAge; or
2. Upgrade packageManager and the documented minimum pnpm version to at least 10.26.0, then validate a clean/frozen install and CI with that version.

If upgrading, note that blockExoticSubdeps is already true by default in supported pnpm versions. Please also document the intended seven-day release-age policy and any exceptions, and update the PR summary so it describes every added setting.

Local read-only validation with pnpm 10.18.3:
- pnpm install --frozen-lockfile --ignore-scripts
- pnpm check

Both pass, but they cannot validate features that the pinned pnpm version does not implement. GitHub CI has not run for this first-time-contributor head (action_required, zero jobs), so it should be executed after the compatibility issue is fixed.

Release decision: No release needed. This changes root development/install policy only; no package changeset is required.

anupamme:
Thanks for the detailed compatibility analysis. Went with option 1: removed blockExoticSubdeps and trustPolicy since they have no effect at the pinned pnpm@10.18.3. minimumReleaseAge: 10080 (7 days) is now the only active setting, which matches the branch scope. A follow-up PR can add trustPolicy once pnpm is upgraded past 10.21.0

Ray-56 (changes requested):
Thanks for removing the two pnpm settings that are unsupported by the repository's pinned pnpm 10.18.3. The supply-chain setting itself is now correctly scoped, and isolated validation confirms that the frozen install, version checks, and package checks pass.

One repository gate still blocks this head:

[P2] Rename the source branch so the enforced branch-name check can pass.

The current head branch is:

`fix-repo-call-e-integrations-pnpm-minimum-release-age`

The repository requires `<type>/<short-kebab-summary>`. Running the exact validator fails with:

```text
Invalid branch name: fix-repo-call-e-integrations-pnpm-minimum-release-age
Expected <type>/<short-kebab-summary> from docs/git-naming-conventions.md.
```

Please rename it to a conforming name such as `fix/pnpm-minimum-release-age`, or push the same commits from a conforming branch if the hosting platform cannot retarget this PR. Then verify:

```bash
pnpm run check:branch-name -- --branch fix/pnpm-minimum-release-age
pnpm run check:versions
pnpm -r --filter "./packages/*" run check
```

[P3] Please also correct the PR metadata before merge. The current title is truncated and does not follow the repository's PR-title convention; `fix(deps): enforce a seven-day pnpm release age` would accurately describe the final change. The Threat Model section says this is a private application that is not published to npm, but this repository publishes multiple `@call-e/*` packages. The accurate scope is that `minimumReleaseAge` hardens dependency resolution for this repository's own development and release installs; downstream consumers still use their own package-manager policy.

GitHub CI remains `action_required` with zero jobs for this first-time-contributor head. After the branch is corrected, a maintainer must allow the workflow and require a green run before merge.

Release decision: No release needed. This is a root dependency-resolution policy change, affects no published package behavior, and needs no changeset.

anupamme:
Closing in favour of a new PR from the correctly named branch `fix/pnpm-minimum-release-age` to pass the repository's branch-name gate.

PR #83, opened 2026-08-12
fix(core): bound the MCP request timeout before arming setTimeout

zkasuran:
## Summary

`@call-e/core` 0.3.0 gave `callMcpTool` a public `timeoutSeconds` override that the CLI passes as 150 for `plan_call`. That value, together with a `config.timeoutSeconds` built by any direct `@call-e/core` consumer, reaches the timer arithmetic in `mcp-client.js` without an upper bound:

```js
Math.max(Math.ceil(Number(timeoutSeconds) * 1000), 1000)
```

`setTimeout` collapses any delay above 2147483647ms (about 24.8 days) to 1ms, so a `timeoutSeconds` above 2147483 aborts the request almost immediately instead of waiting. A non-finite or non-positive value has the same effect, because `Math.max(NaN, 1000)` stays `NaN` and `setTimeout` treats that as 1ms.

This is the immediate-abort failure PR #73 fixed for `--timeout-seconds` at the flag boundary. The CLI validator caps that flag at `MAX_TIMER_SECONDS` and rejects non-finite durations, but the core transport does its own arithmetic, so the new public override and a caller-built config skip that guard.

## Fix

One helper in `packages/core/lib/mcp-client.js` validates the requested seconds before arming either timer:

- The value must be finite, positive and no greater than `MAX_TIMER_SECONDS` (2147483, the largest whole-second delay `setTimeout` keeps intact). An oversized, non-finite or non-positive value is treated as invalid.
- An invalid `config.timeoutSeconds` falls back to the 15s default. An invalid per-call `timeoutSeconds` override falls back to the session timeout the handshake already computed.
- In-range values are unchanged and the existing lower clamp of 1000ms is kept.

Capping an oversized value at the ceiling, which the first version of this fix did, traded the immediate abort for a worse failure: a malformed or attacker-influenced value became a roughly 24.8-day timeout that held the request, socket, AbortController and caller resources open instead of failing. Falling back fails safely.

`openMcpSession` and `callMcpTool` both call the helper, so `config.timeoutSeconds` and the per-call override are validated the same way. No public surface changes: `timeoutSeconds` is still an optional number and the helper is not exported.

## Tests

`packages/core/test/core.test.js` reads the delay handed to `setTimeout` for each request and covers both timer sites:

- an out-of-range per-call override (`MAX_TIMER_SECONDS + 1`), a non-finite one (`Infinity`, `NaN`) and a negative one all fall back to the session timeout
- the same out-of-range, non-finite and negative `config.timeoutSeconds` values fall back to the 15s default for the handshake and the list call
- a per-call override and a session value at `MAX_TIMER_SECONDS` are both accepted unchanged (the boundary)
- an in-range 150s override is passed through unchanged

The two oversized cases fail without the change, since they previously asserted the cap. Reverting only `mcp-client.js` while keeping the tests:

```text
not ok 13 - MCP client falls back to the session timeout for an oversized tool-call override
not ok 19 - MCP client falls back to the default timeout for an oversized session timeout
# tests 30  # pass 28  # fail 2
```

The non-finite and negative cases already fell back under the first version, so they pass either way. The behavior change is that an oversized value now falls back too. Every case has explicit coverage on both the session and per-call paths.

## Verification

From the repository root on Node 22.22.2 with pnpm 10.18.3.

```text
pnpm test           exit 0   core 30/30, cli 58 (57 pass, 1 skip) + 15/15
pnpm check          exit 0   tsc --noEmit clean, syntax check clean
pnpm pack:dry-run   exit 0
```

The single skip is the pre-existing Windows-only case. No CALL-E account or live call is involved; every case drives an injected `fetchImpl`.

## AI assistance

AI (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: `pnpm test`, `pnpm check` and `pnpm pack:dry-run` all pass on Node 22.22.2 with pnpm 10.18.3.

### DIFF
--- a/.changeset/bound-mcp-request-timeout.md
+++ b/.changeset/bound-mcp-request-timeout.md
@@ -1,0 +1,10 @@
+---
+"@call-e/core": patch
+---
+
+Bound the MCP request timeout on both ends before arming setTimeout. A `callMcpTool`
+`timeoutSeconds` override or a `config.timeoutSeconds` above 2147483 seconds used to reach
+`setTimeout` unchanged, where Node collapses any delay past 2147483647ms to 1ms and aborts
+the request almost immediately. Oversized values are now capped at that ceiling. Non-finite
+or non-positive values fall back to the session timeout. This matches the cap the CLI
+already applies to `--timeout-seconds`.
--- a/.changeset/bound-mcp-request-timeout.md
+++ b/.changeset/bound-mcp-request-timeout.md
@@ -2,9 +2,13 @@
 "@call-e/core": patch
 ---
 
-Bound the MCP request timeout on both ends before arming setTimeout. A `callMcpTool`
-`timeoutSeconds` override or a `config.timeoutSeconds` above 2147483 seconds used to reach
-`setTimeout` unchanged, where Node collapses any delay past 2147483647ms to 1ms and aborts
-the request almost immediately. Oversized values are now capped at that ceiling. Non-finite
-or non-positive values fall back to the session timeout. This matches the cap the CLI
-already applies to `--timeout-seconds`.
+Validate the MCP request timeout before arming setTimeout and fall back to a safe value
+rather than capping. A `callMcpTool` `timeoutSeconds` override or a `config.timeoutSeconds`
+reaches Node's timer, where any delay past 2147483647ms (~24.8 days) collapses to 1ms and
+aborts the request almost immediately. The seconds value is now required to be finite,
+positive and no greater than 2147483 (the largest whole-second delay the timer keeps intact).
+An out-of-range or non-finite session value falls back to the 15s default; an out-of-range or
+non-finite per-call override falls back to the already computed session timeout. Capping an
+oversized value at the ceiling was worse than the abort it replaced: it turned a malformed
+value into a ~24.8-day timeout that held the request, socket and caller resources open instead
+of failing. This matches the validation the CLI already applies to `--timeout-seconds`.
--- a/packages/core/lib/mcp-client.js
+++ b/packages/core/lib/mcp-client.js
@@ -2,9 +2,27 @@ import { readJson, tokenCachePath, tokenIsUsable } from "./cache.js";
 import {
   DEFAULT_MCP_CLIENT_NAME,
   DEFAULT_MCP_CLIENT_VERSION,
+  DEFAULT_TIMEOUT_SECONDS,
   INTEGRATION_HEADER,
   MCP_PROTOCOL_VERSION,
 } from "./constants.js";
+
+// setTimeout collapses any delay above 2_147_483_647ms (~24.8 days) to 1ms, so a very
+// large timeout fires the abort almost at once and cancels the request the caller meant
+// to keep waiting on. The CLI already caps --timeout-seconds at this ceiling (see
+// packages/cli/lib/config.js), but the public callMcpTool timeoutSeconds override and a
+// caller-built config.timeoutSeconds reach the timer arithmetic below without that bound,
+// so clamp it here rather than trusting the value.
+const MIN_TIMEOUT_MS = 1000;
+const MAX_TIMEOUT_MS = 2_147_483_647;
+
+function boundedTimeoutMs(seconds, fallbackMs) {
+  const requestedMs = Math.ceil(Number(seconds) * 1000);
+  if (!Number.isFinite(requestedMs) || requestedMs <= 0) {
+    return fallbackMs;
+  }
+  return Math.min(Math.max(requestedMs, MIN_TIMEOUT_MS), MAX_TIMEOUT_MS);
+}
 
 export class AuthRequiredError extends Error {
   constructor(message = "A usable CALL-E auth token is required.") {
@@ -142,7 +160,7 @@ function accessTokenFromCache(config) {
 async function openMcpSession({ config, fetchImpl }) {
   requireFetch(fetchImpl);
   const accessToken = accessTokenFromCache(config);
-  const timeoutMs = Math.max(Math.ceil(Number(config.timeoutSeconds || 15) * 1000), 1000);
+  const timeoutMs = boundedTimeoutMs(config.timeoutSeconds, DEFAULT_TIMEOUT_SECONDS * 1000);
   const commonHeaders = {
     Accept: "application/json, text/event-stream",
     "Content-Type": "application/json",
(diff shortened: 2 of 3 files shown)


Ray-56 (changes requested):
Thanks for addressing the immediate-abort regression and for covering both the session and per-call timer sites. The exact head passes the frozen install, branch-name check, version checks, package checks, and Core tests locally, but one availability issue still blocks merge.

[P2] Treat an oversized timeout as invalid and fall back instead of turning it into a 24.8-day timeout.

`boundedTimeoutMs` currently caps any large finite value at `2_147_483_647ms`. That prevents Node from collapsing the timer to 1ms, but it converts a malformed or attacker-influenced value into an effectively disabled request timeout. A direct `@call-e/core` consumer can supply `config.timeoutSeconds` or the public per-call override without the CLI validator; an unresponsive endpoint can then retain the request, socket, AbortController, and caller resources for about 24.8 days. This replaces immediate failure with resource exhaustion rather than failing safely.

Please validate the seconds value before multiplying it:
- require `Number.isFinite(parsed)`
- require `parsed > 0`
- require `parsed <= Math.floor(2_147_483_647 / 1000)`
- for an invalid session value, fall back to `DEFAULT_TIMEOUT_SECONDS`
- for an invalid per-call override, fall back to the already computed session `timeoutMs`

Keep valid values such as 150 seconds unchanged. Update the tests so `MAX_TIMER_SECONDS + 1`, `Infinity`, and a negative value fall back for both the session and per-call paths; retain a boundary test showing that `MAX_TIMER_SECONDS` itself is accepted. The changeset and PR description should describe fallback rather than capping.

[P3] Please remove the two `APPEND-TIMEOUT-TESTS-*` marker comments from `packages/core/test/core.test.js`; they appear to be generation artifacts and should not ship.

CI is still `action_required` with zero jobs for this first-time-contributor head. Because this changes security-sensitive request timeout behavior, require a real green CI run after the fix rather than merging from local validation alone.

Release decision: Patch release recommended for `@call-e/core`; the existing patch changeset is the correct release shape once the behavior and text are corrected.

zkasuran:
Thanks Ray. Both are fixed and the changeset and PR body now describe fallback rather than capping.

**[P2] Oversized timeout falls back instead of capping.** `boundedTimeoutMs` validates the seconds before multiplying: `Number.isFinite(parsed)`, `parsed > 0` and `parsed <= MAX_TIMER_SECONDS` where `MAX_TIMER_SECONDS = Math.floor(2147483647 / 1000)` is 2147483. An invalid session value falls back to `DEFAULT_TIMEOUT_SECONDS`; an invalid per-call override falls back to the already computed session `timeoutMs`. In-range values such as 150s are untouched and the 1000ms lower clamp stays. A malformed or attacker-influenced value now fails safely rather than becoming a 24.8-day timeout that pins the request, socket and AbortController.

**Tests.** `MAX_TIMER_SECONDS + 1`, `Infinity`, `NaN` and a negative value each fall back on both the session and per-call paths. A boundary test keeps `MAX_TIMER_SECONDS` itself accepted (2147483000ms passes through). Reverting only `mcp-client.js` with the tests in place fails the two oversized cases:

```text
not ok 13 - MCP client falls back to the session timeout for an oversized tool-call override
not ok 19 - MCP client falls back to the default timeout for an oversized session timeout
# tests 30  # pass 28  # fail 2
```

The non-finite and negative cases already fell back before, so they hold either way. The oversized case is the behavior that changed.

**[P3] Marker comments removed.** Both `APPEND-TIMEOUT-TESTS-*` lines are gone from `core.test.js`.

**Changeset and description.** Both now describe validation and fallback, not a cap. Release shape unchanged: patch on `@call-e/core`.

**CI.** Agreed the local run should not stand in for green CI on a security-sensitive change. The checks need your approve-and-run on this first-time-contributor head; I cannot trigger them from the fork. Local run from the repo root on Node 22.22.2 with pnpm 10.18.3:

```text
pnpm test           exit 0   core 30/30, cli 58 (57 pass, 1 skip) + 15/15
pnpm check          exit 0
pnpm pack:dry-run   exit 0
```

Head is now `cbb7eea`.

zkasuran:
All of this is already on the branch. `cbb7eea` was pushed at 03:16Z, 43 minutes after your review at 02:33Z, so the review reads the head before it. Itemising it against your list so you do not have to diff it yourself.

**[P2] Fallback, not capping.** `boundedTimeoutMs(seconds, fallbackMs)` validates before multiplying:

```js
const parsed = Number(seconds);
if (!Number.isFinite(parsed) || parsed <= 0 || parsed > MAX_TIMER_SECONDS) {
  return fallbackMs;
}
return Math.max(Math.ceil(parsed * 1000), MIN_TIMEOUT_MS);
```

`MAX_TIMER_SECONDS` is `Math.floor(2_147_483_647 / 1000)`. The session call site passes `DEFAULT_TIMEOUT_SECONDS * 1000`. The per-call site passes the already computed `timeoutMs`, so an invalid override falls back to the session value rather than to the default. 150 seconds is unchanged.

**Tests, both paths.** Eleven, all running by name: oversized, non-finite, NaN and negative for the tool-call override, the same four for the session timeout, the timer ceiling accepted on both, plus 150 seconds unchanged. `MAX_TIMER_SECONDS` itself asserts `MAX_TIMER_SECONDS * 1000`, so the boundary is exercised rather than assumed.

**[P3]** Both `APPEND-TIMEOUT-TESTS-*` markers are gone. They were generation artifacts and you were right that they should not have shipped.

**Changeset.** Rewritten to describe fallback. It now says the value is required to be finite, positive and no greater than 2147483. It also says capping was worse than the abort it replaced, because it turned a malformed value into a ~24.8-day timeout holding the request, socket and caller resources open.

Ran on `cbb7eea` here: branch-name check passes, `check-manifest-versions` reports metadata in sync, `sync-install-doc-versions --check` reports install references in sync, the Core runtime syntax check passes for 7 files, plus `node --test ./test/*.test.js` in `packages/core` at 30 pass, 0 fail across `core.test.js` and `token-expiry.test.js`. I could not run `check:examples`: its `test_e2e.py` OAuth cases need a live broker and fail here on the network rather than on this diff.

That leaves the one thing you named that is not mine to do. The CI run for `cbb7eea` is still `completed` with conclusion `action_required` and zero jobs, because the head is from a first-time-contributor fork. A fork author cannot release that. If you approve the run the real green result you asked for will exist on the exact head. Happy to rebase onto current `main` first if you would rather the run happened on a fresher base.

PR #99, opened 2026-08-24
feat(hermes-plugin): add Hermes Agent plugin package

Jcube101:
## Summary

Adds `packages/hermes-plugin`, a native plugin for
[[Hermes Agent](https://github.com/NousResearch/hermes-agent)](https://github.com/NousResearch/hermes-agent).

Hermes is listed in the README as a supported client, but its badge reads
`ClawHub Prompt` and the only documented route is the pasted install prompt.
This package gives it the same shape as the other clients: a one-command
install, native tools, and an install doc.

Hermes is OpenClaw's successor and ships a migration path from it, so users
already running the CALL-E OpenClaw skill are the ones most likely to land
here.

```bash
hermes plugins install CALLE-AI/call-e-integrations/packages/hermes-plugin/plugin --enable
```

## What it adds

Five tools in the `calle` toolset — `calle_auth`, `calle_plan`, `calle_run`,
`calle_status`, `calle_show` — over an adapter that drives the `calle` CLI.

- **Planning and dialling are separate tools.** `calle_plan` returns a
  `confirm_summary` for the user and dials nothing; `calle_run` takes the
  `plan_id`. The skill instructs the agent to show the summary and wait.
- **The call authorization never reaches the agent.** `calle_plan` does not
  return `confirm_token`, `calle_run` reads it from local state, and it is
  redacted from the recorded argv. It authorizes a charged call for about a
  day and cannot be revoked early.
- **Person calls disclose.** The opening states that the caller is an AI and
  that the call is transcribed. No flag removes it. Business calls disclose on
  request. The caller may be named per call; unnamed, the disclosure names
  nobody.
- **`goal_returned` is read before dialling**, and a provider-side edit is
  flagged. This is a text-level check: a clause surviving into the plan does
  not predict it will be followed on the call, and nothing here describes it
  as verification.
- **Outcomes are written to local disk.** A run deleted provider-side reports
  as failed, so `calle_show` is the only way to read an older call back.
- **`call start` is not exposed** and no code path builds it. The validator
  asserts this.

## Deliberate divergences from the other client packages

Each of these differs from `claude-plugin` / `codex-plugin` / `cursor-plugin`,
and each is a consequence of how Hermes works rather than a preference.

**No marketplace entry.** `hermes plugins install` resolves a subdirectory
from the identifier itself, so there is no registry to register against and
nothing for a `marketplace.json` to point at. Per
`docs/agent-integration-layout.md`, the reserved identifier would be
`call-e-hermes`; there is no file for it to live in. Nothing is added at the
repository root.

**No sparse checkout in the install doc.** One command covers install and
enable.

**Tools rather than a skill.** Hermes discovers skills from its skills
directory and from `skills.external_dirs`; a plugin's own directory is in
neither, so a `SKILL.md` shipped in a plugin does not load. The agent-facing
surface is `provides_tools`. `plugin/skills/calle/SKILL.md` still ships, as
documentation and for parity, and both it and the install doc say plainly that
it will not appear in `hermes skills list`.

**Python payload.** This is the first package here to ship executable code
rather than markdown alone. That follows from the point above: on the other
clients the skill file is the mechanism, and on Hermes it is not.

## Layout

```text
packages/hermes-plugin/
  plugin/                                  installed by hermes plugins install
    plugin.yaml                            manifest, read at the plugin root
    __init__.py                            register(ctx) entry point
    tools.py                               schemas, handlers, availability check
    adapter.py                             CLI adapter and goal construction
    README.md
    skills/calle/SKILL.md                  documentation; does not load
    skills/calle/references/commands.md
  scripts/check-plugin.mjs
  test/check-plugin.test.js
  package.json  CHANGELOG.md  README.md
docs/install/hermes-plugin.md
```

⚠️ **Only `plugin/` reaches a user.** `hermes plugins install` moves that
directory out of a shallow clone and discards the rest, so `scripts/`, `test/`
and `package.json` are for CI only. `npm pack --dry-run` ships nine files:
`README.md`, `package.json`, and `plugin/**`.

## Validation

```bash
pnpm --filter @call-e/hermes-plugin check
pnpm --filter @call-e/hermes-plugin test
pnpm --filter @call-e/hermes-plugin pack:dry-run
```

`check-plugin.mjs` follows `claude-plugin`'s validator with the constants
retargeted, plus:

- `provides_tools` in the manifest matches the names `tools.py` registers, in
  both directions
- `tools.py` never references `confirm_token`, `cache_path` or `expires_at`
- `adapter.py` builds no `call start` invocation and still redacts the token
- the agent-facing markdown never presents a raw CLI verb as usable

`check-plugin.test.js` has 14 cases, 12 of them negative — each mutates the
package into a specific defect and asserts the validator fails.

Two repository scripts skip this package because they work from fixed lists
rather than globbing: `check-manifest-versions.mjs` looks for
`.claude-plugin/plugin.json` and its siblings, and this package's manifest is
`plugin/plugin.yaml`; `sync-install-doc-versions.mjs` has an explicit
`syncedInstallFiles` list. Both pass as-is. Happy to add this package to either
list if you would rather it were covered — that seemed like your call rather
than mine.

⚠️ One assertion detail worth flagging, since `assertCliGuidance` in the other
packages uses bare `includes()`: a prohibition and its violation are the same
substring. `SKILL.md` names `call start` in order to say it is unavailable, and
`tools.py` has a docstring saying the token never appears in a result. Both
would trip a naive check. The validator here strips Python comments and
docstrings before scanning code, and for prose checks whether a verb is
presented as usable rather than merely mentioned. There is a test for the
allowed case.

## Also changed

- `README.md` — the Hermes badge now reads `CALL-E` rather than
  `ClawHub Prompt`, matching the clients that have a package, and **Agent
  Install** gains the plugin install command. The pasted-prompt paragraph is
  untouched: it remains the route for clients without one. Hermes is already
  listed in the use-case table, so that row needed no change.
- `.changeset/add-hermes-plugin.md` — a `minor` bump for the new package.
- `pnpm-lock.yaml` — two lines registering the workspace member.

## Checked against

Hermes Agent v0.19.0, installed and loaded end to end on a clean instance:
the subdirectory resolved from the `owner/repo/path` shorthand, the plugin
installed to `<hermes home>/plugins/calle`, `--enable` wrote `plugins.enabled`,
and after a gateway restart all five tools registered into the `calle` toolset
with `check_fn` correctly gating them while unauthorized.

Subdirectory support is not visible in `hermes plugins install --help`; it is
carried in the identifier grammar (`owner/repo/path`, `#subdir`, and the
`/tree/<branch>/path` browser URL).

The gateway restart is required after install — tools are registered at gateway
start. The installer prints this, and the install doc and both READMEs say so.

One limitation to be aware of, upstream in Hermes rather than here:
`hermes plugins update` cannot update a plugin installed from a subdirectory,
because the installed copy carries no git metadata. Both the install doc and
the package README give `--force` re-install as the upgrade path. Reporting it
upstream separately.

### DIFF
--- a/.changeset/add-hermes-plugin.md
+++ b/.changeset/add-hermes-plugin.md
@@ -1,0 +1,9 @@
+﻿---
+"@call-e/hermes-plugin": minor
+---
+
+Add the Hermes Agent plugin package. Registers five CALL-E tools (`calle_auth`,
+`calle_plan`, `calle_run`, `calle_status`, `calle_show`) and installs from this
+subdirectory in one command. Planning and dialling are separate tools, calls to
+a person disclose that the caller is an AI, and call outcomes are stored
+locally so they remain readable after the provider deletes its copy.
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ New users get 20 free calls to get started. [Sign up now!](https://www.heycall-e
 ![Claude Code](https://img.shields.io/badge/Claude%20Code-CALL--E-orange)
 ![Cursor](https://img.shields.io/badge/Cursor-CALL--E-blue)
 ![OpenClaw](https://img.shields.io/badge/OpenClaw-ClawHub-purple)
-![Hermes Agent](https://img.shields.io/badge/Hermes%20Agent-ClawHub%20Prompt-green)
+![Hermes Agent](https://img.shields.io/badge/Hermes%20Agent-CALL--E-green)
 ![MCP](https://img.shields.io/badge/MCP-Streamable%20HTTP-blue)
 
 </div>
@@ -125,6 +125,15 @@ Choose the integration path that fits your use case:
 Install CALL-E for me: https://open.heycall-e.com/document/mcp-archive/CALL-E-installation-guide.md
 ```
 Works in Claude Code, Codex, Cursor, and any agent that can run shell commands. The linked guide stays up to date, so the prompt never changes.
+
+**Hermes Agent** installs as a native plugin instead:
+
+```bash
+hermes plugins install CALLE-AI/call-e-integrations/packages/hermes-plugin/plugin --enable
+hermes gateway restart
+```
+
+Registers five CALL-E tools and prompts for sign-in on first use. See the [Hermes plugin install guide](https://github.com/CALLE-AI/call-e-integrations/blob/main/docs/install/hermes-plugin.md).
 
 For manual setup, expand the table below or see the [full install guide](https://github.com/CALLE-AI/call-e-integrations/blob/main/docs/install/install-guide.md).
 
--- a/docs/install/hermes-plugin.md
+++ b/docs/install/hermes-plugin.md
@@ -73,6 +73,16 @@ Planning and dialling are separate tools. `calle_plan` returns a
 `confirm_summary` for the user to read; `calle_run` takes the `plan_id`. The
 call authorization stays in local state and is never returned to the agent.
 
+`calle_run` registers a `pre_tool_call` hook that escalates to Hermes'
+human-approval gate — the same `[o]nce/[s]ession/[a]lways/[d]eny` prompt used
+for dangerous shell commands. The host resolves it at the dispatch site and is
+fail-closed: a denial, a timeout, or an error in the gate all block the call,
+and the tool never runs. The prompt names the callee, the number and the
+purpose, so an `[a]lways` answer does not authorize calls to anyone else.
+
+A plan is single-use: if a run already exists for it, `calle_run` reports that
+run rather than placing a second call.
+
 Calls to a person open by disclosing that the caller is an AI and that the call
 is transcribed. There is no flag to remove it.
 
@@ -123,6 +133,16 @@ pnpm --filter @call-e/hermes-plugin check
 pnpm --filter @call-e/hermes-plugin test
 pnpm --filter @call-e/hermes-plugin pack:dry-run
 ```
+
+The handler tests run against a fake provider and place no calls:
+
+```bash
+python3 packages/hermes-plugin/plugin/test_handlers.py
+```
+
+They assert on side effects rather than return values — that planning submits
+nothing, that a run submits exactly once, and that a retry after a reported
+failure submits nothing at all.
 
 ## More
 
--- a/docs/install/hermes-plugin.md
+++ b/docs/install/hermes-plugin.md
@@ -1,0 +1,109 @@
+# Install The CALL-E Hermes Agent Plugin
+
+The Hermes integration path in this repository is the plugin package at
+`packages/hermes-plugin/plugin`. It registers native Hermes tools and does not
+require a skill file, a marketplace entry, or a pasted prompt.
+
+## Install
+
+```bash
+hermes plugins install CALLE-AI/call-e-integrations/packages/hermes-plugin/plugin --enable
+```
+
+`hermes plugins install` resolves a subdirectory from the identifier, so there
+is no sparse checkout and no marketplace entry point to add. `--enable` turns
+the plugin on without a second command.
+
+## Authorize
+
+In a Hermes conversation:
+
+```text
+sign in to CALL-E
+```
+
+The agent calls `calle_auth`, returns a link, and waits. Open the link, finish
+in the browser, and tell the agent when you are done. No terminal step is
+needed, which matters because most Hermes users reach the agent through
+Telegram, Discord, or Slack rather than a shell.
+
+## CLI Availability
+
+The plugin uses `CALLE_BIN` when set, then a global `calle` command when
+available, then falls back to:
+
+```bash
+npx -y @call-e/cli
+```
+
+Node 22 or newer is required. Set `CALLE_BIN` to pin a specific build.
+
+## Tools
+
+| Tool | |
+| --- | --- |
+| `calle_auth` | Sign in. |
+| `calle_plan` | Build a call plan and return a summary. Dials nothing. |
+| `calle_run` | Place the planned call. |
+| `calle_status` | Poll a call in progress. |
+| `calle_show` | Read a stored plan or outcome from local disk. |
+
+Planning and dialling are separate tools. `calle_plan` returns a
+`confirm_summary` for the user to read; `calle_run` takes the `plan_id`. The
+call authorization stays in local state and is never returned to the agent.
+
+Calls to a person open by disclosing that the caller is an AI and that the call
+is transcribed. There is no flag to remove it.
+
+## Notes
+
+**The skill file does not load.** `plugin/skills/calle/SKILL.md` ships as
+documentation. Hermes discovers skills from its skills directory and from
+`skills.external_dirs` in `config.yaml`; a plugin's own directory is in
+neither, so the file will not appear in `hermes skills list`. The behaviour it
+describes is implemented in the tools.
+
+**Updating.** `hermes plugins update` cannot update a plugin installed from a
+subdirectory: the installer moves that directory out of a shallow clone, so the
+installed copy carries no git metadata and the update command exits with an
+error. Re-run the install command with `--force` instead.
+
+```bash
+hermes plugins install CALLE-AI/call-e-integrations/packages/hermes-plugin/plugin --force --enable
+```
+
+**Configuration is optional.** Call outcomes and the token cache live under
+`<hermes home>/plugins/calle`. See
+[packages/hermes-plugin/plugin/README.md](../../packages/hermes-plugin/plugin/README.md)
+for the environment variables, including the optional `CALL_CONSENT_CMD` hook
+for sites that keep a record of who has agreed to be called.
+
+## Local Development
+
+From a clone of this repository:
+
+```bash
+hermes plugins install file:///path/to/call-e-integrations#packages/hermes-plugin/plugin --force --enable
+```
+
+The `#subdir` fragment works for any scheme. Restart the Hermes gateway so the
+new plugin directory is picked up, then check it loaded:
+
+```bash
+hermes plugins list
+hermes tools
+```
+
+## Verify The Package
+
+```bash
+pnpm --filter @call-e/hermes-plugin check
+pnpm --filter @call-e/hermes-plugin test
+pnpm --filter @call-e/hermes-plugin pack:dry-run
+```
+
+## More
+
+See [packages/hermes-plugin/README.md](../../packages/hermes-plugin/README.md)
+for package layout, how this package differs from the other clients, and
+validation commands.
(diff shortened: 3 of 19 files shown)


Ray-56 (changes requested):
Requesting changes for head `0d351aea502f7cb21c980a40004ba63321f99c8b`.

**P1 — `calle_run` can place a real call and then report failure.** `tools.py` builds `Namespace(plan_id=..., max_wait=...)` without `wait`; `adapter.cmd_run()` submits `provider.run()` first and only then reads `args.wait`. A fake-provider reproduction recorded one submission followed by `AttributeError: Namespace has no attribute wait`. The handler catches that as `calle_run failed`, so an agent or user can retry and produce a duplicate charged call. `calle_status` has the same missing field and fails before polling. Set and validate `wait`/`max_wait` explicitly, and add runtime handler tests that assert exactly one submission and successful status polling.

**P1 — approval and uncertain-submission safety are instructions, not enforcement.** The bundled `SKILL.md` explicitly does not load, and hiding `confirm_token` does not stop the model from calling `calle_run` immediately or after prompt injection from a callee transcript. Register a fail-closed Hermes `pre_tool_call` approval directive for `calle_run` so denial, timeout, or gate failure cannot reach `provider.run`. Also make plan execution single-use: if local state already has a `run_id`, return or poll that run instead of submitting again. Preserve and consume the CLI recovery information for an uncertain `run_call` submission rather than converting it to a generic error that invites a retry. Add tests for denial, timeout, repeated tool calls, stable `run_id` reuse, and uncertain recovery.

Provider/callee text is also returned to model context in `raw`, summaries, and transcripts, while the only data-not-instruction rule lives in the unloaded documentation. Remove provider instruction fields and unnecessary `raw` data from model-visible results, and put the untrusted-output rule in a runtime-loaded system prompt section or equivalent host-enforced surface.

**P2 — the release path is currently broken.** In an isolated exact-head checkout, `pnpm run version-packages` changed `@call-e/hermes-plugin` to `0.2.0` but left `plugin/plugin.yaml`, `adapter.py` `__version__`, and the CLI attribution docs at `0.1.0`; the package check then failed on the manifest mismatch. Extend the repository version sync/check scripts to cover the Hermes YAML manifest, runtime attribution, and shipped reference files, and add a release-versioning test. The minor release classification is reasonable for the new capability, but the package is not releasable until these surfaces stay synchronized.

**P2 — runtime coverage is missing several observable defects.** `calle_show` reads only `<id>.json`, while terminal outcomes are written to `<run_id>.result.json`, so it cannot provide the persisted outcome promised by the docs. The local `VALID_REGIONS` set also rejects many regions currently documented by this repository. Load the result sidecar for run IDs, avoid a second stale region source of truth, and add Python behavior tests for plan/run/status/show and every currently supported region.

**P2 — protect the spend credential at file creation time.** The plan sidecar contains `confirm_token`, but the state directory uses default permissions and the fixed `.tmp` file is written under the process umask before a later `chmod(0600)`. Create the directory as `0700`, create atomic temporary files as `0600` without following a pre-existing symlink, then replace the destination; test the resulting modes.

CI for this head is `action_required` with zero jobs. The package validator, its 14 tests, Python compilation, and package dry-run pass locally, but none of those tests execute the handlers, which is why the post-submission failure is currently missed.

Jcube101:
Thanks — this was a thorough review and the first P1 was a real defect. I reproduced it with a fake provider: one submission, then AttributeError on wait, and a retry placing a second charged call.

P1 — duplicate submission. tools.py built the args object from what the tool accepts rather than what the verb reads, and the provider is invoked before wait is read. Args are now declared per verb and checked against the adapter at import, so a field the adapter reads but the handler doesn't set fails at load rather than after a call. A plan is single-use — if a run exists, calle_run polls it instead of submitting. An uncertain outcome returns the run_id and says not to re-run.

P1 — approval. You were right that the mechanism exists; I'd concluded it didn't. pre_tool_call now escalates calle_run through request_tool_approval, fail-closed on denial, timeout or gate error. No rule_key, so the derived key comes from a message naming this callee and number — [a]lways can't authorise a call to anyone else. A plan whose recipients can't be read returns block rather than asking for approval of a call it can't describe.

P2 — release path. adapter.py reads its version from plugin.yaml. Your scenario reproduced: bumping package.json alone now fails the check, and the runtime follows the manifest with no third edit.

P2 — calle_show. Reads the result sidecar and resolves a plan id to its run. While fixing it I found cmd_run was writing the carry state under both ids, so confirm_token existed in two files — the run-keyed copy now omits it.

P2 — regions. Removed the local list rather than syncing it. A copy of the provider's regions can only go stale, and it fails in the wrong direction by refusing a call that would have worked. Format check only.

P2 — permissions. Directory 0700 at creation, mkstemp plus chmod before the rename. POSIX modes are advisory on Windows, so that assertion skips there with a printed reason rather than passing for the wrong one.

On coverage — you're right, and it's the part I'd most want to take from this. Fourteen tests, twelve of them negative, and not one executed a handler; they asserted properties of files. plugin/test_handlers.py now runs the tools against a fake provider and asserts on submission counts, which is what catches a call that dials and then reports failure. It found two further bugs in my own fixes while I was writing it.

The workflow is awaiting maintainer approval, so CI hasn't run on any head yet. Happy to rebase or split this if that helps.

Ray-56:
Thanks for the substantial work here, especially the follow-up on approval enforcement and duplicate-submission safety.

Before we can justify adding and maintaining a dedicated Hermes package, we need a clearer product and architectural case for why it is **necessary**, rather than an optional native integration.

## Current baseline

In our current validation, CALL-E already works in Hermes through the existing OpenClaw/ClawHub route and through the portable `skills/calle` skill. Hermes v0.19.0 also supports direct GitHub skill installation, so the existing skill can be installed with a Hermes-native one-command path:

```bash
hermes skills install CALLE-AI/call-e-integrations/skills/calle
```

That skill and the shared `calle` CLI already cover authentication, MCP tool discovery, call planning/execution, uncertain-submission recovery, status polling, terminal results, and the untrusted-output boundary. Therefore, the question is not whether Hermes users can use CALL-E today; they can. The question is which mandatory requirements cannot be satisfied by the existing skill plus shared CLI and justify a second runtime implementation.

Could you please expand the PR rationale with concrete answers to the following?

## 1. What exact user-facing gap requires a plugin?

Please identify specific workflows that fail, are unsafe, or are materially unreliable with `skills/calle` in Hermes. For each gap, it would help to include:

- a reproduction using a supported Hermes version;
- the observed behavior with the existing skill;
- the required behavior;
- why a small Hermes-specific skill adjustment or shared CLI change cannot provide it.

“Same shape as the other clients,” “native tools,” and a different badge are useful UX goals, but they are not by themselves reasons that a separate runtime is required.

## 2. Is host-enforced approval a product requirement or an optional hardened mode?

The clearest unique capability in this PR is the Hermes `pre_tool_call` hook that fail-closes `calle_run` through the host approval gate. A Markdown skill cannot provide an equivalent host-level guarantee.

Please clarify:

- Is a second, host-enforced approval immediately before every real call now a CALL-E requirement?
- If it is mandatory, why should this behavior exist only in Hermes while the current OpenClaw and portable skill flows treat clear initial user intent as sufficient?
- If it is not mandatory across clients, should this package be described as an optional “hardened native integration” rather than the required Hermes support path?
- Which threat model or observed failure demonstrates that the existing `call start` flow and CLI recovery protections are insufficient?

This is an important product-policy decision, not merely a client packaging difference.

## 3. Why does the Hermes layer own client-independent behavior?

The package currently implements or reimplements substantial behavior around:

- single-use plan execution and uncertain-submission handling;
- confirmation-token persistence;
- result persistence and read-back;
- consent and disclosure rules;
- goal construction and field extraction;
- status and transcript normalization;
- region/language validation and polling behavior.

Most of these concerns are not Hermes-specific. Please explain why they should live in a Python adapter used by one client instead of being added to `@call-e/cli`, where OpenClaw, skills.sh, Codex, Claude Code, Cursor, and Hermes can share the same tested behavior.

If native Hermes tools are still valuable, could the shared CLI own these semantics while the plugin remains a thin layer containing only tool schemas, availability checks, and the approval hook?

## 4. How will the duplicated runtime remain in parity?

This PR adds about 3,500 lines, including roughly 1,250 lines in `adapter.py` and 670 lines in `tools.py`. For comparison, the existing shared CLI implementation under `packages/cli/lib` is about 2,180 lines in total.

The first review already found examples of the risks created by this second execution layer: post-submission failure, duplicate charged-call risk, stale region validation, result-sidecar mismatch, release-version drift, and new local credential-file requirements. The follow-up addresses several of those defects, but it also demonstrates the ongoing maintenance surface.

Please describe the long-term parity plan for:

- CLI commands, arguments, and output shapes;
- recovery and idempotency behavior;
- supported regions and terminal statuses;
- auth/cache changes;
- safety and disclosure policy changes;
- release version synchronization;
- integration and compatibility testing against supported Hermes releases.

Who will own these checks when either Hermes or `@call-e/cli` changes?

## 5. Why is the plugin lifecycle a net improvement?

The existing skill can be installed directly by Hermes, participates in Hermes skill discovery/security scanning, and has a tracked update lifecycle. The proposed subdirectory plugin:

- requires a gateway restart;
- cannot currently use `hermes plugins update` from its installed subdirectory;
- requires a forced reinstall for upgrades;
- adds a second local state store containing call authorization and result data.

Please compare the complete install, upgrade, rollback, and troubleshooting experience and explain why the plugin is materially better for users despite those costs.

## 6. Please revisit the plugin-skill loading claim

The PR says a plugin-shipped `SKILL.md` cannot load because the plugin directory is not a normal skill directory. However, Hermes v0.19.0 documents `ctx.register_skill()` for registering namespaced plugin skills. The current `register(ctx)` implementation registers tools and the approval hook, but does not register the shipped skill.

Please clarify the supported Hermes version range and whether there is a compatibility reason not to use `ctx.register_skill()`. If the file is intentionally documentation-only, please explain why that is preferable to directly installing or registering the existing portable skill.

## 7. What evidence will determine whether the plugin is worth keeping?

It would help to define measurable acceptance criteria, for example:

- fewer incorrect tool/argument selections than the skill-based flow;
- a demonstrated unsafe call path that only the host hook prevents;
- materially better authentication or recovery completion rates;
- acceptable install/update friction;
- compatibility results across the supported Hermes versions and platforms.

Without concrete criteria, we risk accepting a large parallel implementation because it feels more native, while taking on continuing safety and release risk without knowing whether users benefit.

## Suggested decision boundary

Our current view is:

- **Baseline Hermes support:** reuse `skills/calle` plus the shared `calle` CLI, with a small Hermes install guide and any necessary attribution/wording adjustments.
- **Dedicated Hermes plugin:** justified only for explicit host-enforced capabilities that a skill and shared CLI cannot provide, and preferably implemented as a thin adapter over shared CLI behavior.

We are open to a dedicated plugin, but the PR needs to demonstrate why the additional Hermes-only runtime is required and why its benefits outweigh the duplicated safety, release, state, and maintenance surface.

Ray-56:
Closing this for now. Hermes can already use CALL-E through the existing OpenClaw/ClawHub route and the portable skills/calle skill backed by the shared calle CLI, so we do not currently have enough justification for maintaining a separate Hermes runtime of this size. We can reconsider or reopen this if there is concrete evidence that a required host-enforced capability cannot be provided by the existing skill and shared CLI. A smaller follow-up focused on Hermes installation documentation, attribution, or a thin approval-only integration would also be welcome.

PR #106, opened 2026-09-04
fix(cli): emit the JSON error envelope for transport failures

N-45div:
### Summary

A failing `calle auth login` currently prints a bare string to stderr and writes **nothing to
stdout**, so an agent host has no JSON to parse and no `error.code` to branch on.

`runCli` rethrows every error that is not an `InvalidArgumentsError`, so HTTP and transport
failures skip `writeCommandError` entirely and land in `main()`, which only prints
`error.message`. That contradicts the contract the skill sets for agent hosts, in
`skills/calle/references/commands.md`:

> Treat all command output as JSON except `--help`.
> If `ok` is false and `error.code` is `auth_required`, run or suggest `auth login`, then retry.

An agent following that instruction gets an empty stdout on any network-layer failure.

Separately, `HttpStatusError` already captures `statusCode` and `responseText`, but nothing
reads them, so the upstream error body is discarded.

### How I hit this

Fresh install on Windows, Node v24.11.1, `@call-e/cli` 0.5.0, clean token cache:

```bash
npm install -g @call-e/cli
calle auth status --json          # usable: false, no cache
calle auth login --start-only --no-browser-open --json
```

**Before**

```
Client error '502 Bad Gateway' for url 'https://seleven-mcp-sg.airudder.com/api/v1/openagent-auth/sessions'
```

Nothing on stdout. The reason is only visible if you curl the broker by hand:

```json
{"error":"oauth_register_failed",
 "message":"Failed to register an OAuth client. err_type=HTTPStatusError"}
```

The rest of the estate was healthy at that moment — the MCP endpoint returned 401 and
`api.heycall-e.com` returned 405 — so the natural reading of a bare 502 is "my install is
broken", and the documented recovery path (reinstall, re-auth) cannot help.

**After**, against the same live 502:

```json
{
  "ok": false,
  "server_url": "https://seleven-mcp-sg.airudder.com/mcp/openagent_oauth",
  "error": {
    "code": "oauth_register_failed",
    "message": "Client error '502 Bad Gateway' for url '.../sessions' Failed to register an OAuth client. err_type=HTTPStatusError The CALL-E login service is unavailable. This is not a local configuration problem, so reinstalling the CLI will not help. Retry later, or use the Developer API with a dashboard API key, which does not depend on brokered login.",
    "status_code": 502,
    "remote_error": {
      "code": "oauth_register_failed",
      "message": "Failed to register an OAuth client. err_type=HTTPStatusError"
    }
  }
}
```

### Changes

- `runCli` routes **every** failure through `writeCommandError`, not just
  `InvalidArgumentsError`. The `help_command` hint stays exclusive to argument errors, so that
  path is byte-identical to before.
- `errorPayload` gains an `HttpStatusError` branch carrying `status_code`, the parsed upstream
  `remote_error`, and a 500-character bounded fallback when the body is not JSON (gateways
  often return HTML).
- A 5xx specifically from the brokered-login registration endpoint adds one sentence saying the
  failure is server-side and that the Developer API path does not depend on it.

### Testing

- `packages/cli`: 55 unit + 15 e2e passing, including two new cases — a JSON upstream error
  body, and a long non-JSON body asserting the bound holds.
- Also ran the suites for `core`, `claude-plugin`, `codex-plugin`, `cursor-plugin`,
  `openclaw-cli-skill`, `skills-sh-skill`: 142 tests total across the repo, 0 failures.
- `node scripts/check-runtime-syntax.mjs` passes for all 8 CLI files.
- Verified by hand against the live broker while it was returning 502.
- I could not run `pnpm pack:dry-run` in my environment; the change touches no package
  metadata.

### Notes

Changeset included. Happy to split the "login service is unavailable" wording into a separate
commit if you would rather keep the envelope fix purely mechanical.


Ray-56 (changes requested):
[P1] Do not promote an untrusted upstream error body directly into the CLI's trusted error contract or terminal output.

`parseRemoteErrorBody` bounds only non-JSON text. A JSON response can supply arbitrarily long `code` and `message` strings; the upstream `code` becomes the top-level `error.code`, while the message is duplicated in the JSON envelope and written raw to stderr. A compromised, misconfigured, or user-selected broker can therefore spoof stable local codes such as `auth_required`/`invalid_arguments`, inject CR/LF or ANSI terminal controls into logs, amplify output, and expose internal response details that the CLI previously withheld.

Please keep the top-level code locally owned (`broker_unavailable` or `http_error`) and retain a normalized upstream code only under `remote_error`. Reuse one remote-string sanitizer for both JSON and non-JSON bodies: allowlist fields, cap code/message lengths, constrain machine codes to a safe character set, strip or escape terminal controls before stderr, and avoid surfacing token-like/internal fields. Add hostile JSON tests covering a long flat and nested message, forged local code, CR/LF/ANSI content, and a secret-like field that must not appear in stdout or stderr.

[P2] The central transport guarantee and documentation are incomplete. Both new tests return HTTP responses; neither covers `fetch` rejecting with `TypeError`/DNS/connection failure. Add a genuine rejected-fetch regression asserting nonzero exit, parseable stdout JSON, no `help_command`, and bounded/safe stderr. Update the canonical `packages/cli/docs/cli-reference.md` and synchronized `packages/cli/README.md`, which still state that some top-level failures may print plain stderr, and explicitly document the stable envelope fields.

CI for `359ee27` is `action_required` with zero jobs. The exact head passes `pnpm check`, `pnpm test`, and `pnpm pack:dry-run` in an isolated checkout, but this touches authentication/error handling and cannot be merged without a real green required CI run after the fix.

Release decision: **Patch release recommended** for `@call-e/cli`; the changeset targets the correct package and bump, but its claim that upstream bodies are safely bounded must be corrected with the implementation.

N-45div:
Thanks for the review — both points were right, and P1 was a real hole: I had promoted whatever `code` the upstream body sent straight into the CLI's own `error.code`. Addressed in the follow-up commit.

**P1 — trusted contract**

- `error.code` is now always CLI-owned: `broker_unavailable` (5xx from the brokered-login registration endpoint), `http_error` (any other non-success status), or `transport_error` (fetch rejected / timed out before a response). Nothing from an upstream body can reach the top-level code, so `auth_required` and `invalid_arguments` cannot be spoofed.
- Upstream detail lives only under `error.remote_error`, allowlisted to `code` and `message` (top-level or nested under `error`). Every other field is dropped unread, so token-like values never touch stdout or stderr.
- One sanitizer for both JSON and non-JSON bodies: I hardened the existing `safeRemoteString` rather than adding a second helper, so MCP call errors get the same treatment. It strips ANSI CSI/OSC/ESC sequences and C0/C1 controls, then bounds (500 chars for messages, 300 for the local status line). `safeRemoteCode` constrains machine codes to `[A-Za-z0-9_.:-]{1,64}` and drops anything else rather than "cleaning" it into something plausible. `writeCommandError` strips controls once more before stderr as a last line of defence.
- Hostile tests added: forged `auth_required` (asserts no `login_url`/`assistant_hint` leak through), 20 KB flat message, nested error object with internal `details`/`request_id` fields, CR/LF/ANSI content, `access_token`/`token`/`refresh_token` fields absent from stdout **and** stderr, and an unsafe code (`bad code[31m`) dropped.

**P2 — transport coverage and docs**

- Added a genuine rejected-fetch regression (`TypeError("fetch failed")` with `cause.code = ENOTFOUND`): exit 1, parseable stdout, `error.code = transport_error`, `cause_code = ENOTFOUND`, no `help_command`, stderr bounded and control-free. Plus a request-timeout case classified the same way.
- `cli-reference.md` gains an **Error Envelopes** section documenting `ok`, `server_url`, `error.code` (with the full list of CLI-owned values and exit codes), `error.message`, `error.status_code`, `error.remote_error`, `error.cause_code`, and `help_command`. Both it and `README.md` no longer say some failures print plain stderr.
- Changeset corrected to describe the sanitization instead of claiming bodies are "bounded".

**Testing**

- `packages/cli`: 67 unit + 15 e2e green (12 new). Other packages: 72 green. `check-runtime-syntax`, branch-name, manifest-version and install-doc sync all pass locally.
- Also ran the patched CLI against the live broker again while writing this; it is answering normally now, so I could not re-reproduce the 502 end to end — the hostile-body tests cover that path.

On CI: I can't trigger the workflow from the fork side (`action_required`, first-time contributor), so it needs an approval click from a maintainer whenever you're ready. Happy to split the docs into a separate commit if that's easier to review.

Ray-56 (changes requested):
Requesting changes on the current head (`f57ed62`). The broker-body spoofing and rejected-fetch coverage from the previous review are improved, but the new global error-envelope path still has untrusted-output gaps:

- **[P1] Sanitize every remote MCP error path, not only `HttpStatusError`.** Direct `mcp tools` / `mcp call` JSON-RPC errors still place the server-controlled `McpHttpError.message` into `error.message` without a bound or sanitizer. On this exact head, a local hostile MCP fixture produced a 20,022-character parsed message and copied a secret-like value to both stdout and stderr; the parsed message also retained ESC. `call start` similarly interpolates an unbounded remote clarification question into `plan_not_ready`. Centralize the remote-message boundary (preferably where `McpHttpError` is created), bound and strip controls before constructing all public envelopes, and add hostile MCP-error and clarification tests covering long text, CR/LF/ANSI, and secret-like content.
- **[P1] Keep upstream HTTP detail out of the trusted top-level summary/stderr.** The PR says upstream detail lives only under `remote_error`, but `remoteError.message` is appended to top-level `error.message` and automatically printed to stderr. A token or internal value embedded inside the allowlisted message is therefore still exposed. Keep the top-level/stderr summary locally authored; retain bounded, clearly untrusted detail only in `remote_error`, with redaction/tests for token-like material inside the message itself.
- **[P2] Use a typed transport boundary.** `isTransportFailure` classifies every JavaScript `TypeError` as `transport_error`, although the documented contract says that code means a request received no response. An unrelated local bug can therefore be presented as a retryable-looking network condition. Have the HTTP layer throw a dedicated transport/timeout error and test an unrelated `TypeError`.
- **[P2] Synchronize the published contract and telemetry.** The documented “full” code list omits existing call/recovery codes, and MCP/call timeouts do not consistently use `transport_error`. `errorTelemetryCode` also records the new broker/transport outcomes as `local_error`, while the README duplicates the canonical contract without a `sync-with` marker. Document the actual matrix, centralize code classification, and add telemetry/contract tests.

The exact head passes `pnpm check`, `pnpm test`, and `pnpm pack:dry-run` in an isolated checkout using the repository-pinned pnpm 10.18.3. GitHub CI remains `action_required` with zero jobs; because this changes authentication/error handling, it cannot use the zero-job low-risk exception. A real required CI run must pass after the fix.

Release decision: **Patch release recommended** for `@call-e/cli`. The existing patch changeset targets the correct package, but its security and stable-code claims must be corrected to match the final implementation.

N-45div:
Thanks — all four points were correct, and the two P1s were the same mistake in two places: I had built a boundary at the HTTP-body path and left the MCP JSON-RPC path and the clarification interpolation untouched, and I was still copying upstream text into the trusted summary. Round 2 moves the boundary to where errors are created, so there is nowhere left for remote text to enter an envelope unsanitized.

**[P1] Every remote MCP error path**

- New `@call-e/core/sanitize` — `stripTerminalControls`, `redactSecrets`, `safeRemoteString`, `safeRemoteCode`, `sanitizeRemoteError` — imported by both the core MCP client and the CLI. One implementation.
- `McpHttpError.message` is now always locally authored (`Remote MCP error for tools/list`). The server's text stays raw in `payload` (existing tests still pass) and sanitized in the new `remoteError` field, which is the only thing the CLI ever displays, under `error.remote_error`.
- `plan_not_ready` no longer interpolates the clarifying question. The summary is fixed text; the question is under `error.remote_error.message`, sanitized and bounded.
- Tests: hostile `tools/list` and `tools/call` JSON-RPC errors (20 KB, CR/LF/ANSI, bearer token, `access_token=…`, a `refresh_token` in `error.data`) — none of it reaches stdout or stderr; the parsed message has no ESC; a hostile clarifying question gets the same treatment.

**[P1] Upstream detail out of the trusted summary**

- The HTTP-failure summary is now built from the status code and our own request URL (`HTTP 502 from https://…/api/v1/openagent-auth/sessions.`) plus the local hint — never the response's status text or body. stderr prints only that.
- `redactSecrets` also runs *inside* allowlisted messages: bearer/basic credentials, `token=` / `secret=` / `password=` style pairs, well-known prefixes (`sk_`, `ghp_`, `xox…`, `AKIA`, …), and long opaque hex/base64 runs. Tested with secrets embedded in message text, not just in fields.

**[P2] Typed transport boundary**

- `core/http.js` throws `TransportError` when `fetch` rejects or the request times out (`url`, `method`, `timedOut`, `code` = `timeout` or the system error code found via `causeCodeOf`, which walks `cause` because `fetch failed` nests it).
- The MCP client marks timeouts and rejected fetches as `McpHttpError` with `code: "transport_error"`, `transport`, `timedOut`, `causeCode`. Timeouts were previously `http_error`; that core test is updated.
- The CLI sets `transport_error` / `error.transport: true` only from those types. The `TypeError` heuristic and the message-regex timeout detection are gone. New test: a `TypeError` thrown from inside the CLI's own polling loop is `internal_error` with no `transport` field.

**[P2] Contract and telemetry**

- `ERROR_CODES` is exported and `classifyError` is the single mapping used by the envelope, stderr, and `errorTelemetryCode`. `local_error` is gone; broker and transport outcomes report their real codes (test covers `auth_login_local_failed` → `broker_unavailable` and `mcp_tools_checked` → `transport_error`).
- `cli-reference.md` now documents the complete matrix: the six general codes plus `plan_not_ready`, `plan_call_invalid_response`, `run_call_missing_run_id`, `recovery_not_found`, `recovery_storage_error`, `{plan_call,run_call,get_call_run}_{error,timeout}`, and `internal_error`, with exit codes and the `transport` / `remote_error` / `cause_code` fields. A test parses that table and asserts it equals `Object.keys(ERROR_CODES)` with matching exit codes, so the docs cannot drift from the code.
- README has a `<!-- sync-with: docs/cli-reference.md#error-envelopes -->` marker and no longer duplicates the list.
- Changeset now covers `@call-e/core` too and describes the actual mechanism.

**Testing**

- core: 23 green (3 new). cli: 73 unit + 15 e2e green. Other packages: 51 green. `check-runtime-syntax` on 18 files, branch-name, manifest-version, install-doc sync all pass. `check:types` passes against the updated `.d.ts` files.

CI still shows `action_required` from the fork side — it needs a maintainer approval to run. I'm happy to split core and cli into two commits if that helps the release.

Ray-56 (changes requested):
Requesting changes on the exact current head (8f08207). The previous head improved the shared error contract, but the following issues remain:

- **[P1] Keep remote call-result text out of the trusted summary and stderr.** `callCallStage` still constructs `CallStageError` with `remoteError.message` for `CallToolResult.isError` and missing-`run_id` paths. `errorPayload` then publishes that server-controlled text as top-level `error.message` and stderr, contradicting the CLI reference and changeset promise that these fields are CLI-authored. Use fixed local summaries and retain sanitized remote detail only in `error.remote_error`; update the tests that currently expect the remote message at the top level.
- **[P1] Control characters can split credentials before redaction.** `safeRemoteString` strips terminal controls by inserting a space before applying credential patterns. For example, `access_token=abcd<ESC>[31m1234efgh` becomes `access_token=[redacted] 1234efgh`, and a similarly split `sk_live_` token remains nearly intact. Canonicalize or ignore control sequences while detecting secrets, then sanitize output. Add adversarial CSI/OSC/C0 insertion tests for bearer/basic credentials, key/value secrets, prefixed tokens, and opaque tokens, asserting that stdout, stderr, and `remote_error` contain no recoverable fragments.
- **[P2] Body-read failures escape the typed transport boundary.** The new catches wrap `fetch` but not `response.text()` in both core HTTP helpers. A timeout or socket reset after headers therefore escapes as a raw `AbortError`, producing incorrect envelope, telemetry, and retry diagnostics. Map body-stream failures to `TransportError`/`McpHttpError` and test slow or rejected bodies for generic HTTP and MCP, especially real-call stages.
- **[P2] Make the documented error contract match the implementation.** Numeric remote codes bypass `safeRemoteCode` (`1e100` becomes the prohibited `1e+100`); `remote_error` can contain undocumented stage fields; and rejected call-stage transports emit a `*_error` code with `transport:true` although the table defines those codes as non-transport. Apply one validation/classification path and add semantic parity tests, not only code-name/exit-code checks. Also update the canonical example, which still embeds upstream text in top-level `error.message`.
- **[P2] Correct the core release decision and public API documentation.** This head adds root exports and the new public `@call-e/core/sanitize` package subpath. Repository release guidance classifies a new package entry point/additive public API as a **minor** core release, not patch. Change the core changeset to minor and document/type-test the public subpath and new error APIs in `packages/core/README.md`; keep the API internal instead if it is not meant to be supported.

The exact head passes `pnpm check`, `pnpm test`, and `pnpm pack:dry-run` in an isolated checkout using pnpm 10.18.3. GitHub reports zero CI jobs. Because this changes authentication/error handling and public package surfaces, it is excluded from the low-risk zero-job exception and needs a real green required CI run after the fixes.

N-45div:
Thanks — all five held up, and two of them I should have caught myself. Addressed on the new head.

**[P1] Call-stage summaries.** The `isError` path and the missing-`run_id` path now use fixed local text (`run_call returned an error.`, `run_call did not return a run_id.`). The server's `message` and `error_code` ride under `error.remote_error` only. The two tests that expected the remote message at top level are updated to expect the local summary and to find the remote text under `remote_error`.

**[P1] Split credentials.** You were right that stripping with a space was the bug. `stripTerminalControls` now *removes* control sequences and control characters (CSI, OSC, two-character ESC, C0/DEL/C1, including CR/LF) with no replacement, so `access_token=abcd<ESC>[31m1234efgh` canonicalizes to `access_token=abcd1234efgh` and is redacted as one credential. Redaction runs on that canonical text; bounding happens last. New core test inserts seven different sequences (CSI colour, CSI erase+home, OSC hyperlink, ESC M, NUL, CR LF, a C1 byte) at four positions — inside the key name, right after the separator, mid-value, and at offset 3 — into bearer, basic, key/value, `sk_live_`, `ghp_`, and opaque-hex credentials, and asserts neither half survives. A CLI test does the same through a hostile broker body and checks stdout, stderr, and `remote_error`. The cost is that legitimate line breaks are joined rather than spaced; I took security over cosmetics there and documented it.

**[P2] Body-read failures.** Both `requestJson` and `requestJsonRpc` now wrap `response.text()`: an `AbortError` after headers is a timeout (`TransportError` / `McpHttpError` with `timedOut`), anything else is a transport failure carrying the cause code. Tests cover both helpers with an aborted body and an `ECONNRESET` body, and a real call stage (`call start` → `plan_call`) for each: `plan_call_timeout` with `transport: true, cause_code: "timeout"`, and `transport_error` with stage fields and `cause_code: "ECONNRESET"`.

**[P2] Contract parity.**
- A rejected/reset transport inside a stage is now `transport_error` (keeping `stage`, `call_started`, `retry_safe`), not `<stage>_error` with a contradictory flag. `classifyError` reads `transport` from the `ERROR_CODES` table only; it is never inferred from the error object, so the code and the flag cannot disagree.
- Numeric remote codes go through `safeRemoteCode`: safe integers only (`-32000` kept, `1e100` / `1.5` / `NaN` dropped). The charset is documented as `-?[A-Za-z0-9][A-Za-z0-9_.:-]{0,63}` and the leading minus is the only change from before.
- `publicRemoteError` is the single validator for every `remote_error` emission: at most `{ code, message }`, each validated, nothing else. Stage fields cannot appear there.
- New parity test runs ten scenarios (invalid args, broker 5xx/4xx, broker fetch rejected/timeout, MCP JSON-RPC error, MCP fetch rejected, `plan_not_ready`, stage `isError`, stage body reset) and asserts for each: code is in the table, exit code matches, `Boolean(error.transport) === ERROR_CODES[code].transport`, the summary and stderr contain no remote marker, and `remote_error` has only `code`/`message` with a valid code charset. The docs table test from last round still runs.
- The canonical example in `cli-reference.md` now shows a CLI-authored `error.message`.

**[P2] Release and public API.** Core changeset is now **minor**. `packages/core/README.md` lists the `@call-e/core/sanitize` subpath and gains an "Errors and Remote Text" section documenting `TransportError`, `causeCodeOf`, the new `McpHttpError` fields (`remoteError`, `transport`, `timedOut`, `causeCode`), and each sanitize function. `test/types.ts` imports and exercises all of them, so `check:types` covers the public subpath.

Same CI caveat as before: `action_required` needs a maintainer approval to run. Happy to split core and cli into two commits, or the sanitizer into its own PR, if that makes the minor release cleaner.

Ray-56 (changes requested):
Requesting changes on the exact current head (6c206597). The previous findings are substantially improved, but the remote-text boundary is still bypassable:

- **[P1] Canonicalize complete C1 sequences and Unicode format controls before redaction.** The sanitizer recognizes ESC-prefixed CSI/OSC, then removes C1 bytes individually. A valid C1 OSC inserted inside a sensitive key, `access_to\u009d8;;x\u0007ken=abcd1234efgh5678`, becomes `access_to8;;xken=abcd1234efgh5678` and leaks the credential. Zero-width format characters also split bearer, prefixed, and opaque tokens while remaining visually invisible; a 40-character opaque value split with U+200B survives intact. Remove complete 7-bit and 8-bit CSI/OSC sequences plus Unicode default-ignorable/bidi format controls before secret detection. Test every insertion position with C1 CSI/OSC, U+200B/U+2060, bidi isolates/overrides, and newlines, asserting no recoverable fragments in stdout, stderr, or `remote_error`.
- **[P1] Keep invalid successful-response bodies out of the trusted summary.** `requestJson` lets `JSON.parse(text)` throw a native `SyntaxError`; Node includes remote input in messages such as `Unexpected token R, "REMOTE-TEXT-MARKER" is not valid JSON`. The CLI generic branch then emits that remote excerpt as top-level `error.message` and stderr, contradicting the locally-authored guarantee and leaking credential prefixes. Convert malformed 2xx JSON into a typed invalid-response error with a fixed local summary, retaining only sanitized/bounded detail under `remote_error`; add hostile invalid-JSON tests.
- **[P2] Apply machine-code validation consistently.** `safeRemoteCallError` still validates `error_code` with `safeRemoteString` or accepts any number, then republishes it as `error.error_code`. Use `safeRemoteCode` with the same safe-integer rule, or remove the duplicate field, and add contract tests for long, fractional, exponential, whitespace, and punctuation-bearing codes.
- **[P2] Align the public transport contract with body-read failures.** Runtime now classifies a reset after headers as transport, but the CLI field table, core JS JSDoc, and HTTP/MCP declarations still say transport means that no HTTP response was received or only a fetch rejection. The CLI also says a body-read reset failed before a response was received. Use a neutral no-usable-response summary or preserve the failure phase, and synchronize README/reference/JSDoc/declarations and semantic tests.

The exact head passes `pnpm check`, `pnpm test`, and `pnpm pack:dry-run` locally with pnpm 10.18.3. GitHub CI is still `action_required` with zero jobs; because this changes auth/error handling and public package APIs, the low-risk zero-job exception does not apply. The core minor plus CLI patch changeset is otherwise correct.

N-45div:
All four are real, and I reproduced the two P1s before touching anything. Fixed on the new head.

**[P1] 8-bit introducers and invisible format characters.** Confirmed exactly as described. `access_to<U+009D>8;;x<BEL>ken=abcd1234efgh5678` stripped to `access_to8;;xken=abcd1234efgh5678` — the credential fully intact, because removing a lone U+009D leaves its parameters behind as ordinary text and the key name no longer matches. A 40-character opaque token split with U+200B survived untouched, and U+2060 inside a bearer credential leaked the tail half.

Removal is now by *sequence*: CSI and OSC are recognised in both their `ESC [` / `ESC ]` and their 8-bit `U+009B` / `U+009D` forms, with BEL, `ESC \` and 8-bit ST all accepted as OSC terminators, and `\p{Cf}` plus `\p{Default_Ignorable_Code_Point}` cover the invisible characters.

That fix alone caused a regression worth describing, because it is not obvious. A CSI sequence consumes its final byte, and an attacker can pick a final byte that belongs to the word being searched for: `Bea<U+009B>rer abcdef…` is a *valid* CSI sequence terminating in `r`, so correct stripping yields `Beaer` — the keyword stops matching and the credential leaks. My own adversarial test caught it. Terminal-accurate stripping and secret detection genuinely want opposite things here, so detection now runs over **two canonicalizations** — sequence-consuming and character-only — and when one reading finds a credential the other does not, the entire string is replaced with `[redacted]` rather than displayed. Text that hides a credential in control codes does not get shown.

Tests cover 5 credential shapes × 12 insertions (8-bit CSI, both OSC terminators, 7-bit CSI, ZWSP, word joiner, ZWJ, RLO, LRI, soft hyphen, BOM, CRLF) at every third offset, asserting no recoverable fragment in stdout, stderr, or `remote_error`.

**[P1] Invalid successful bodies.** Also confirmed: `JSON.parse` puts its input into its own message, so a `text/plain` 200 produced `Unexpected token R, "REMOTE-TEXT-MARKER…" is not valid JSON` as top-level `error.message` *and* stderr. `requestJson` now catches it and throws a typed `InvalidResponseError` carrying the raw body in `responseText` and never in `message`. The CLI publishes it as a new `invalid_response` code with a fixed local summary and sanitized detail under `remote_error`. Two things fell out: arrays were being accepted as JSON objects (`typeof [] === "object"`), now rejected; and the hostile test asserts no `sk_live_` prefix reaches either stream.

**[P2] `error_code` validation.** Now goes through `safeRemoteCode` with the same safe-integer rule, and `status` with it too. One refinement on your wording: a code carrying control characters is dropped outright rather than accepted after cleaning — `bad<ESC>[31m` had been surviving as the entirely plausible `bad`, which is the "clean it into something valid" behaviour the docs say we refuse. Contract tests cover long, fractional, exponential, whitespace-bearing, space-bearing and control-bearing codes.

**[P2] Transport contract.** `TransportError` and `McpHttpError` carry `phase` — `connect` when nothing arrived, `body` when the stream failed after headers — and the CLI message names the phase instead of claiming nothing was received. The reference table, the field list, `packages/core/README.md`, the JSDoc and the `.d.ts` declarations all now say "no usable response", with `error.phase` documented and semantically tested.

**Two more, which I found by attacking the round-4 head myself rather than waiting for you to** (`3e75a95`):

- **The wholesale-redaction fallback was too narrow, and leaked.** It only fired when the displayed reading redacted *nothing*, so a message carrying two credentials — one visible in each canonical reading — satisfied that test on the strength of the first and published the second in full: `access_token=[redacted] and Beaer abcdefghijklmnopqrstuvwxyz012345`. The readings are now compared by how many credentials each finds, so the alternate seeing *more* is what triggers the fallback, regardless of what the displayed reading already caught. Regression test covers the two-secret case, and asserts ordinary prose is not over-redacted — an over-eager sanitizer that redacts everything is its own failure.
- **`McpHttpError` carried no `phase`,** while the reference table I added in the same commit promised `error.phase` on transport errors. Both MCP transport paths now tag `connect` or `body`, the CLI emits it, and a non-transport error reports `null` rather than a misleading default.

**Testing:** core 31 (6 new), cli 79 unit + 15 e2e (2 new), other packages 51. `check-runtime-syntax` on 18 files, `check:types` against the updated declarations, branch-name, manifest-version and install-doc sync all pass.

CI is still `action_required` from the fork — it needs a maintainer approval to run. Happy to split the sanitizer into its own PR if that would make the core minor easier to release on its own.

Ray-56 (changes requested):
Requesting changes on the exact current head (`3e75a950`). The previous round fixed the reported single-form gaps, but two P1 boundary failures remain:

- **[P1] Equal redaction counts can still leak a credential found only by the alternate canonical form.** `safeRemoteString` compares only the number of `[redacted]` markers and keeps the display form on a tie. With one credential detectable only in each form, both counts are one but the display output still contains the alternate-only Bearer token. I reproduced this on this head with an 8-bit OSC inside `access_token` plus an 8-bit CSI inside `Bearer`; the result was `access_token=[redacted] and Beaer abcdefghijklmnopqrstuvwxyz012345`. Fail closed whenever the canonical forms produce different credential findings/results, or carry source-position taint and redact the union of matched spans. Add crossed equal-count and multi-secret adversarial tests, not only the current alternate-count-greater case.

- **[P1] Call-stage body failures lose their transport phase and can mislead retry decisions.** `callStageErrorFrom` does not propagate `McpHttpError.phase`, and `CallStageError` does not accept it, so a body-stream timeout/reset is reconstructed with the default `phase: connect` and a message claiming the failure happened before a response. This contradicts the published `connect` versus `body` contract and can encourage an unsafe retry of a real call. Preserve `phase`, use phase-aware or neutral local summaries, and assert `phase: body` for plan/run body timeouts and resets. Synchronize the stale no-response wording in the core JSDoc and declarations.

This head also conflicts with merged #115/#119 across CLI code, tests, docs, and metadata. Resolve against current `main` while preserving both `lib/sanitize.js` and `scripts/run-agent-command.mjs` in syntax checks, the verified shell-free launcher contract, and per-command attribution flags, then rerun the complete suite.

The exact head passes `pnpm check`, `pnpm test`, and `pnpm pack:dry-run` in an isolated checkout with pinned pnpm 10.18.3. GitHub CI is still `action_required` with zero jobs; authentication/public-API changes cannot use the low-risk exception and require a real green run after the fixes.

Release decision: **Minor release recommended** for `@call-e/core` and **patch release recommended** for `@call-e/cli`; the current changeset is otherwise aligned.

Ray-56 (changes requested):
Requesting changes on the exact current head (`4aed14f`). The merge conflict and transport-phase findings are fixed, but the revised sanitizer and merged documentation still have blocking security problems:

- **[P1] Comparing matched strings still misses crossed identical credentials.** `secretFindings` has no occurrence or source-position identity. If one canonical form finds the first copy of a credential and the other finds an identical second copy, the sorted finding strings compare equal and the display form leaks the second copy. On this head, a display-only `Bearer abcdefghijklmnopqrstuvwxyz012345` made with 8-bit OSC plus an alternate-only identical Bearer made with 8-bit CSI returns `Bearer [redacted] and Beaer abcdefghijklmnopqrstuvwxyz012345`. Track source occurrences and redact the union, or fail closed whenever the canonical forms differ and either form finds a credential. Add repeated-identical crossed and multi-copy tests.

- **[P1] Other terminal string controls and unterminated sequences still bypass redaction.** The sequence regex consumes CSI and terminated OSC only. DCS, SOS, PM, and APC are stripped as lone introducer/terminator bytes while their payload remains; unterminated OSC is handled the same way. I reproduced both 7-bit and 8-bit DCS: `access_to<U+0090>junk<U+009C>ken=abcd1234efgh5678` becomes `access_tojunkken=abcd1234efgh5678`, leaving the credential unchanged. Consume every 7-bit/8-bit terminal string-control sequence through its terminator, and fail closed or consume through end-of-input when unterminated. Test all four families, OSC, both forms, terminators, and every credential insertion position through the public helper and CLI envelope.

- **[P1] The added error-contract documentation revives unsafe command execution.** `packages/cli/docs/cli-reference.md` lines 164-165 and 181 say to run `next_command` and call `help_command` directly runnable, contradicting the same canonical page lines 76-81, which require shell-free `*_argv` execution and mark `*_command` display-only. These command strings use POSIX quoting and can carry returned IDs, URLs, or paths; executing them is unsafe and nonportable in PowerShell/cmd. Document `next_argv`/`help_argv` as the only executable follow-ups, keep command strings display-only, and add a schema/docs regression.

- **[P2] A literal NUL makes the new security boundary look binary to Git.** `packages/core/lib/sanitize.js` uses a literal NUL in the findings delimiter; `git diff --numstat` reports `- -`, hiding the entire 10.5 KB sanitizer from normal review. Use an escaped textual source representation such as `join("\\0")` or a length-prefixed serialization. Also synchronize the P3 README/type wording for findings disagreement and body-phase cause codes.

The exact head passes `pnpm check`, core tests (32/32), CLI unit tests (80 pass, 1 platform skip), and `pnpm pack:dry-run`. The full suite reaches one existing base-branch mixed-install test failure on this macOS host because a local Bash `npm` wrapper is invoked with Node; this is not in the PR diff. GitHub CI remains `action_required` with zero jobs, and auth/public-API changes require a real green run after fixes.

Release decision: **Minor release recommended** for `@call-e/core` and **patch release recommended** for `@call-e/cli`; the current changeset levels are correct.

Ray-56 (changes requested):
Requesting changes on the exact current head (`74de30a`). The previous repeated-secret, string-control, command-documentation, NUL, and transport-phase findings are fixed, but two new boundary gaps remain:

- **[P1] Mixed terminal sequences still bypass both global canonicalizations.** `safeRemoteString` computes only “consume every sequence” and “remove every control byte” forms. A message that needs a different choice for different sequences can evade both. On this head, `access_to<OSC8>junk<ST8>k<CSI8>en=abcd1234efgh5678` produces `access_tokn=abcd1234efgh5678`; the all-sequence form loses the `e`, the character-only form keeps both payloads, neither detects `access_token`, and the credential is published intact. Fail closed for ambiguous controlled text (the simplest safe rule is to withhold it whenever the canonical forms differ), or track source-position taint and redact the union across per-sequence interpretations. Add mixed OSC/CSI/DCS multi-sequence tests through both `safeRemoteString` and CLI stdout/stderr/`remote_error`.

- **[P2] The new JSON-RPC outcome check accepts a response for another request.** `requestJsonRpc` checks only for an own `result` or object `error`; it does not require `jsonrpc: "2.0"`, match the response `id` to the request, or reject a body containing both outcomes. A local exact-head fixture returned `{ "jsonrpc": "1.0", "id": "stale-request", "result": { ... } }` and `listMcpTools` accepted it. Require the protocol version, exact request-ID correlation, and exactly one well-formed `result`/`error` before treating the response as successful. Add wrong/stale ID, wrong version, and both-fields regressions; preserve empty notification acknowledgements. This overlaps the protocol-hardening work tracked in #110.

The exact head merges cleanly with current `main`; `pnpm check`, core tests, CLI unit tests, and `pnpm pack:dry-run` pass. The full suite reaches the known base-branch macOS mixed-install fixture failure because the local Bash `npm` wrapper is invoked with Node, not because of this diff. GitHub CI is still `action_required` with zero jobs, so auth/public-API code requires a real green required run after the fixes; the existing `CHANGES_REQUESTED` state also prevents merging.

Release decision: **Minor release recommended** for `@call-e/core` and **patch release recommended** for `@call-e/cli`; the changeset levels remain correct.

Ray-56 (changes requested):
Requesting changes on the exact current head (`194bad3`). The prior mixed terminal-sequence redaction bypass and JSON-RPC version/ID/outcome validation findings are fixed, but the public HTTP contract now disagrees with runtime behavior:

- **[P2] `requestJson` still permits array response types in TypeScript while the implementation rejects every array.** `packages/core/lib/http.d.ts` declares `requestJson<T extends object>`, and arrays satisfy `object`, so existing callers can compile `requestJson<string[]>()`. On this head, `packages/core/lib/http.js` rejects that previously accepted 2xx JSON array as `InvalidResponseError`. Align the contract before release: either preserve array responses, or constrain and document the API as a JSON record/object and add a negative type test. If the published array-capable contract is intentionally removed, reassess the semver level rather than shipping it as a minor release.

The exact head passes `pnpm check`, core tests (41/41), CLI unit tests (86 pass, 1 platform skip), and `pnpm pack:dry-run`. The full suite reaches the known base-branch macOS mixed-install fixture failure because a local Bash `npm` wrapper is invoked with Node; this is not in the PR diff. GitHub CI remains `action_required` with zero jobs, and this auth/public-API change requires a real green required run after the fix.

Release decision: the included core-minor/CLI-patch changeset is otherwise aligned, but a deliberately breaking `requestJson` type/runtime contract needs a major release or compatibility preservation.

N-45div:
Addressed the Sep 14 P2 on the current head in `2db0e35`.

- Preserved the published `requestJson<T extends object>` compatibility by accepting JSON arrays at runtime again; no breaking type/API restriction.
- Added a runtime regression proving an array response round-trips unchanged.
- Added a public type consumer for `requestJson<Array<{ id: string }>>`.
- Synchronized `packages/core/README.md` and the existing changeset; removed the stale "arrays are no longer accepted" claim.
- Merged current upstream `main` first, preserving the released package versions.

Validation on Windows:
- `pnpm check` passes.
- Core: 42/42 tests pass.
- CLI unit: 87/87 tests pass.
- All other package tests pass.
- CLI e2e: all 34 available `powershell.exe` / `cmd` and non-shell cases pass. Six duplicate `pwsh` cases cannot start because PowerShell 7 is not installed on this host (`ENOENT`); the equivalent Windows PowerShell cases pass.
- Manual package dry-runs for both `@call-e/core` and `@call-e/cli` pass. The root CLI pack script is POSIX-only (`mktemp`/`trap`) on Windows, so I verified the same `pnpm pack --pack-destination` operation with a PowerShell-created temporary directory.

Release decision remains the existing core minor + CLI patch changeset. Preserving arrays avoids the major-release path.

@Ray-56 please re-review when ready.

Issue #111, opened 2026-09-05
Packaged agent commands fail in Windows PowerShell because Unix env syntax is mandatory


PionneerZ:
The packaged Codex, Claude Code, Cursor, OpenClaw, and skills.sh instructions require every CLI invocation to start with Unix `env NAME=value ...` syntax. Stock Windows PowerShell has no `env` executable, so the command fails before Node, `calle`, or `npx` starts.

Tested at commit `e497e1a01e171f9559f87e0f998318a9efc06609` on Windows PowerShell 7 with Node.js v24.12.0.

## Reproduction

Run the first documented Codex command literally:

```powershell
Get-Command env -ErrorAction SilentlyContinue
env CALLE_SOURCE=codex CALLE_INTEGRATION=codex_plugin CALLE_INTEGRATION_VERSION=0.1.11 node packages/cli/bin/calle.js --version
```

Actual result:

```text
Get-Command env: NOT FOUND
env: The term 'env' is not recognized as a name of a cmdlet, function,
script file, or executable program.
```

Changing the tail to either `calle --help` or `npx -y @call-e/cli --help` gives the same pre-execution failure because every fallback retains the shared `env` prefix.

Control experiment using native PowerShell syntax:

```powershell
$env:CALLE_SOURCE = 'codex'
$env:CALLE_INTEGRATION = 'codex_plugin'
$env:CALLE_INTEGRATION_VERSION = '0.1.11'
node packages/cli/bin/calle.js --version
```

Actual: `0.5.0`, exit code 0.

## Scope

At this commit there are 145 command lines beginning with `env CALLE_SOURCE` under `packages/`, `skills/`, and `docs/`. The incompatible mandatory pattern appears in:

- `packages/codex-plugin/plugin/skills/calle/SKILL.md:56-80`
- `packages/codex-plugin/plugin/skills/calle/references/commands.md`
- `packages/claude-plugin/plugin/skills/calle/SKILL.md`
- `packages/cursor-plugin/plugin/skills/calle/SKILL.md`
- `packages/openclaw-cli-skill/skills/phone-call-calle/SKILL.md`
- `skills/calle/SKILL.md`
- `docs/install/CALL-E-installation-guide.md`

This blocks `--help`, auth, MCP readiness, planning, execution, and status on a normal Windows PowerShell host. The agent instructions also declare the attribution prefix mandatory, so a conforming agent cannot safely omit it.

## Suggested fix

[...]

PR #115, opened 2026-09-07
fix: run agent commands through a verified MCP launcher

### LINKED ISSUE #109: @call-e/cli and @call-e/calle both install a 'calle' binary, and the plugin skills prescribe the bare command
`@call-e/cli` and `@call-e/calle` both publish a binary called `calle`, and they are not the same program.

```
@call-e/cli     bin: { "calle": "./bin/calle.js" }
@call-e/calle   bin: { "calle": "./dist/cli.js" }
```

One drives the Developer API and has `calls create`, `calls get`, `goals run`, keyed off `CALLE_API_KEY` against `api.heycall-e.com`. The other is the MCP and brokered-login tool with `auth login`, `call plan`, `call run`, `call recover`, defaulting to a different host entirely. Different commands, different auth, different service.

Install both and npm links one of them without a word:

```bash
npm install @call-e/calle@0.7.0 @call-e/cli@0.5.0
cat node_modules/.bin/calle | grep -o "@call-e/[a-z-]*/[a-z/.]*"
# @call-e/calle/dist/cli.js
```

The SDK wins. No warning, no error, nothing in the install output.

That would be a footnote if the plugins did not prescribe the bare command. `packages/claude-plugin/plugin/skills/calle/references/commands.md` offers three forms, and the middle one is plain `calle`:

```bash
env CALLE_SOURCE=claude ... node packages/cli/bin/calle.js call run --plan-id <plan_id> --confirm-token <confirm_token>
env CALLE_SOURCE=claude ... calle call run --plan-id <plan_id> --confirm-token <confirm_token>
env CALLE_SOURCE=claude ... npx -y @call-e/cli call run --plan-id <plan_id> --confirm-token <confirm_token>
```

In a project that has the SDK installed, the middle one runs the wrong program:

```
$ calle call run --plan-id p --confirm-token t
Unknown option: --plan-id
exit 1
```

An agent following the skill gets that back mid-workflow, with a plan already created and a confirm token in hand. The message does not hint that a different `calle` answered. The first and third forms are fine, so which of the three the agent happens to pick decides whether the call goes out.

Installing the SDK to write server code and the CLI to drive agent flows is a normal thing to want, and both are the shipped path for their own use case.

## Suggested fix

Cheapest is to drop the bare `calle` form from the plugin references and keep the two unambiguous ones. `npx -y @call-e/cli` names the package, so it cannot be shadowed, and the explicit `node packages/cli/bin/calle.js` path is fine for a checkout.

The longer answer is one of the two binaries getting a different name. That is a breaking change, so it may not be worth it, but two published packages fighting over one command name will keep producing this.

JJasonSun:
Agent skills can select the SDK's shared `calle` binary, and interpolating user text or opaque call IDs into shell commands can execute or corrupt those values. The five skills now use a bundled launcher that verifies a trusted `@call-e/cli` package and its command help before forwarding a JSON argument array with `shell: false`.

The launcher accepts a request file or stdin and applies integration attribution through the child process environment. The same documented command runs in Bash, PowerShell, and cmd. CLI responses include `login_argv`, `help_argv`, and `next_argv` for exact follow-up execution; existing command-string fields remain available for display. Packaged skills, installation guides, CLI references, and version checks follow this contract.

Fixes #109.

Validation:
- [Linux and Windows CI passed](https://github.com/CALLE-AI/call-e-integrations/actions/runs/34206072356) at `4cf11b6ffe1b3a2ed9c67a278f1c2956739e7add`. Linux passed 169 JavaScript tests, with one Windows-only skip, plus 8 Python OAuth tests. Windows passed 100 CLI tests with no skips.
- Executed 67 documented argument arrays in Bash and all 67 in each of PowerShell 7, Windows PowerShell, and cmd. Verified help, authentication/readiness, call/recovery flows, attribution, and Unicode paths against a local broker/MCP server.
- Installed `@call-e/calle@0.7.0` and `@call-e/cli@0.5.0` together, reproduced npm selecting the SDK, and exercised the candidate CLI source in that installation. Wrong package identity and incompatible help fail before request arguments are forwarded; neither the SDK nor an earlier `calle` on PATH receives authentication or confirmation arguments.
- Adversarial quotes, command substitutions, backticks, backslashes, newlines, and Chinese text reach the fake server unchanged. Lost-response recovery accepts one logical call, and no injected command runs.
- `pnpm check`, `pnpm pack:dry-run`, `git diff --check`, and isolated release-version synchronization passed. Test calls use synthetic credentials and a local server.

Includes patch changesets for `@call-e/cli`, `@call-e/codex-plugin`, `@call-e/claude-plugin`, and `@call-e/cursor-plugin`. OpenClaw and skills.sh receive the synchronized launcher and guidance through their existing repository distribution.


Ray-56 (changes requested):
Requesting changes on the exact current head (826905b7). The package-selection direction is safer than a bare executable, but the agent-facing command contract is still unsafe and incomplete:

- **[P1] Do not interpolate dynamic values into shell command templates.** The synchronized command references place the latest user message verbatim inside single-quoted JSON and place remote `plan_id`, `confirm_token`, `run_id`, and recovery values inside double quotes. A U+0027 quote can break the JSON/shell boundary, while `$()` and backticks are still executable inside double quotes. This can turn an ordinary call request or server-returned opaque value into local command execution or argument corruption before the verified CLI starts. Require shell-free argv execution for every dynamic value, using `JSON.stringify` plus `spawn`/`execFile` with `shell:false`, or add a file/stdin input that avoids shell interpolation. Add adversarial quote, command-substitution, backtick, backslash, and newline tests and assert that no injected process runs and exact arguments reach only the verified entry.
- **[P1] The new mandatory entry syntax still blocks Windows PowerShell.** The packaged skills continue to prefix commands with POSIX `env NAME=value`, and the newly expanded standalone guidance uses `node "$CALLE_CLI_ENTRY"`; stock PowerShell has no `env` command and reads environment variables as `$env:CALLE_CLI_ENTRY`. This is already reproduced in #111, and this PR replaces previously runnable bare CLI examples with the incompatible form. Provide a cross-platform launcher or explicit Bash, PowerShell, and cmd forms selected before execution, fail closed for unknown shells, and add a Windows CI job that runs literal help/auth-status/no-call readiness commands while preserving attribution.
- **[P2] Automate the acceptance cases required by #109.** The E2E test reads the repository copy of `@call-e/cli` and shadows PATH with a fake `calle`, but it does not install/exercise `@call-e/calle` alongside it and has no runtime wrong-package-identity or incompatible-help fail-closed case. Add those fixtures and assert that no auth or confirmation argument reaches either the conflicting SDK binary or a mismatched executable.

The exact head has green GitHub CI and passes `pnpm check`, `pnpm test`, and `pnpm pack:dry-run` locally with pnpm 10.18.3, but the P1 findings block merge. Release decision: patch changesets for `@call-e/cli` and the three published plugins are appropriate; the private OpenClaw/skills.sh validation packages need no changeset.

Issue #122, opened 2026-09-08
Resumed PENDING login polls without displaying or reopening the authorization URL


kutterdev:
# Resuming a pending broker login silently polls without showing its login URL

Tested at `2f438c35f99398219d23e3ff9a1e83edc4cebd76` on Windows with Node.js 24.13.0. This is an offline reproduction with an injected broker response and synthetic cache, not a claim about a live CALL-E account.

If `calle auth login` is interrupted before browser authorization, the cached session can remain `PENDING`. Rerunning the normal login command reuses that session but neither prints its login URL nor invokes the browser opener. A user who closed the original browser tab or lost the original terminal output sees a polling wait without the action needed to finish authorization.

In `packages/core/lib/broker-client.js`, `loginWithBroker` wraps both the URL output and `openBrowser` in `if (created)`. `ensurePendingLogin` correctly reconciles and returns an existing active session with `created: false`; that reuse also suppresses all authorization guidance. The CLI's normal login path calls this helper directly.

I ran the same pending broker response through two cases:

| Case | Requests before first sleep | URL printed | Browser opener invoked |
| --- | --- | --- | --- |
| Fresh session | POST, GET | Yes | Yes |
| Cached PENDING session | GET, GET | No | No |

The test stops at the first sleep using an injected sentinel exception, so it does not wait for a real timeout. No provider requests, actual credentials, browser launch, or calls were involved. The existing 21 core tests also pass.

Minimal reproduction from the repository root (save as an `.mjs` file):

```js
import fs from 'node:fs';
import os from 'node:os';
import path from 'node:path';
import { loginWithBroker } from './packages/core/lib/broker-client.js';
import { pendingCachePath, writePrivateJson } from './packages/core/lib/cache.js';

const cacheRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'calle-resume-'));
const config = {
  cacheRoot,
  serverUrl: 'https://broker.example/mcp/test',
  brokerBaseUrl: 'https://broker.example',
  timeoutSeconds: 1,
  pollTimeoutSeconds: 1,
  minTtlSeconds: 0,
};
const session = {
  session_id: 'synthetic-session',
  session_secret: 'synthetic-secret',
  login_url: 'https://broker.example/resume',
  status: 'PENDING',
  created_at: new Date().toISOString(),
  expires_at: '2030-01-01T00:00:00Z',
};
const printed = [];
const opened = [];
const stop = new Error('Stop after first pending poll');
[...]

PR #125, opened 2026-09-10
fix(auth): validate broker sessions across clients

mauricemohr88-debug:
Malformed broker responses or cached pending sessions could reach caches, login guidance, browser openers, request paths, and session-secret headers. This change validates the session fields before those operations in the shared core, the three advertised broker examples, and CLI status/auth-required output.

Login URLs must match a configured broker or auth origin at sensitive operations; HTTPS is required except for an exact configured loopback HTTP origin. IDs remain bounded opaque strings, including Unicode and separators, and are encoded as one path segment in status/exchange requests. Dot-only IDs, controls, malformed Unicode, and unsafe header secrets are rejected. Invalid cached sessions cannot supply a displayed login URL or a broker request.

The existing one-argument `normalizePendingSession(payload)` contract continues to accept safe custom-origin payloads. Its optional `BrokerOriginPolicy` only contains origin settings, without a request-timeout requirement. Normalization without a policy checks structure and safe schemes; it does not establish origin trust. Core effectful callers always provide their configuration.

Validation includes the required workspace checks/tests/package dry runs, runtime and type compatibility cases, hostile response/cache regressions, and all three broker example suites. CLI regressions additionally exercise read-only status and authorization guidance using independently malformed fields. Tests use local fixtures with external network and private-file access blocked; no real CALL-E accounts or phone calls are used. The Windows-only command-shell test remains skipped on macOS. Required hosted CI must still be approved and pass for this head; local results are not a substitute.

Release decision: patch releases recommended for `@call-e/core` and `@call-e/cli`; the changeset covers both. Package versions are unchanged. The core README, canonical CLI reference, and broker-example READMEs describe the resulting behavior. Prepared with Codex AI assistance and independently reviewed by another Codex agent; no human QA is claimed.

### DIFF
--- a/.changeset/tidy-brokers-validate.md
+++ b/.changeset/tidy-brokers-validate.md
@@ -1,0 +1,5 @@
+---
+"@call-e/core": patch
+---
+
+Reject incomplete broker login session responses before persisting pending authentication state.
--- a/.changeset/tidy-brokers-validate.md
+++ b/.changeset/tidy-brokers-validate.md
@@ -2,4 +2,5 @@
 "@call-e/core": patch
 ---
 
-Reject incomplete broker login session responses before persisting pending authentication state.
+Validate broker session IDs, secrets, and login URLs for their cache, request,
+header, and browser-opening sinks before persisting pending authentication state.
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -87,6 +87,13 @@ Operator-facing integrations can use `ensurePendingLogin` when they need to
 show the login URL and its remaining lifetime without blocking on the full
 login flow.
 
+Broker session fields are validated before they reach the pending-login cache,
+browser opener, request path, or session-secret header. HTTPS login URLs must
+use the configured `brokerBaseUrl` or `authBaseUrl` origin. HTTP is accepted
+only when the URL is a loopback address on that exact configured origin, which
+keeps local development support without letting a remote broker redirect a
+client to an arbitrary local service.
+
 ## Outbound Call Contract
 
 CALL-E's outbound tools follow this order:
--- a/packages/core/lib/broker-client.d.ts
+++ b/packages/core/lib/broker-client.d.ts
@@ -2,6 +2,7 @@ import type { JsonObject, PendingLoginDocument, TokenDocument } from "./cache.js
 
 export interface BrokerRequestConfig {
   brokerBaseUrl: string;
+  authBaseUrl?: string;
   timeoutSeconds: number;
   integrationHeader?: string;
 }
@@ -85,7 +86,10 @@ export function exchangeBrokerSession(
   options?: BrokerRequestOptions,
 ): Promise<TokenDocument>;
 
-export function normalizePendingSession(sessionPayload: BrokerSessionPayload): PendingLoginDocument;
+export function normalizePendingSession(
+  sessionPayload: BrokerSessionPayload,
+  config?: BrokerRequestConfig,
+): PendingLoginDocument;
 
 export function ensurePendingLogin(
   config: BrokerLoginConfig,
--- a/packages/core/lib/broker-client.js
+++ b/packages/core/lib/broker-client.js
@@ -1,5 +1,5 @@
 import { pendingCachePath, pendingIsExpired, readPendingLogin, removeFile, tokenCachePath, tokenIsUsable, writePrivateJson, readJson } from "./cache.js";
-import { INTEGRATION_HEADER, SESSION_SECRET_HEADER } from "./constants.js";
+import { DEFAULT_BASE_URL, INTEGRATION_HEADER, SESSION_SECRET_HEADER } from "./constants.js";
 import { HttpStatusError, requestJson } from "./http.js";
 
 function integrationHeaders(config) {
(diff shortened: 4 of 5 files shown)


Ray-56 (changes requested):
Requesting changes on the exact current head (`a1d014a`).

**[P1] Validate broker session fields for their security-sensitive sinks, not only for presence.** `requiredSessionField` accepts every nonblank string unchanged. I verified this head accepts `login_url: "javascript:alert(1)"` and an OSC-bearing URL, a path-shaping `session_id: "../exchange?x="`, and `session_secret: "ok\r\nX-Evil: 1"`. Those values are then cached; `login_url` is printed and passed to the OS URL opener, `session_id` is interpolated into broker request paths, and `session_secret` becomes an HTTP header. A compromised/misconfigured or user-selected broker can therefore trigger an arbitrary registered URL handler or terminal-control output, reshape follow-up requests, or turn header parsing into a denial of login. The new tests cover only missing/null/number/blank values and do not exercise any sink.

Please validate each field according to its use before writing the pending cache: parse `login_url`, require an allowed HTTP(S) scheme and the repository's intended trusted broker/auth origin policy, reject credentials/control characters and unreasonable lengths; constrain and bound the opaque session id and URL-encode it when building paths; reject control characters and unreasonable lengths in the header value. Add regressions proving hostile values are neither cached, printed/opened, nor sent in a request/header. Coordinate the trusted-origin behavior with #122 and the overlapping URL-hardening work in #47 rather than introducing a conflicting policy.

GitHub CI for this head is `action_required` with zero jobs. Because this changes authentication and token/session handling, the low-risk zero-job exception does not apply; a real required CI run must pass after the fix.

Release decision: **Patch release recommended** for `@call-e/core`; the included patch changeset is otherwise appropriate.

mauricemohr88-debug:
Addressed the requested broker-field validation and regression coverage in the updated head. The trusted origins come from the existing broker/auth configuration; the HTTP loopback exception remains bound to an explicitly configured origin. This preserves the development direction of #47 and pending-session reuse from #122. Unsafe cached sessions, dot-only path IDs, terminal controls, and non-header-safe secrets are covered as well.

Full local tests, checks, package dry-runs, and all 8 Python example tests passed in isolation. The Windows-only test remains unverified locally. The patch changeset and PR description are updated.

@Ray-56, could you review the updated head and approve its required GitHub CI run? My contributor account has no upstream CI approval permission. I will address any actual job failures once the jobs can run; I am not treating the current zero-job state as a pass.

Ray-56 (changes requested):
Requesting changes on the exact current head (`18bc3c8`). The previous scheme/origin, path, header, cache, browser, and control-character sink findings are fixed inside `@call-e/core`, but two compatibility/scope gaps remain:

- **[P2] The public normalization API now breaks previously valid custom-origin and opaque-ID callers without declaring that break.** `normalizePendingSession(payload, config?)` keeps `config` optional, but omitting it silently restricts `login_url` to the default CALL-E origin; a one-argument custom-broker payload that worked on `main` now throws. The new ID allowlist also rejects safe opaque values even though the request path is bounded and `encodeURIComponent`-encoded. Preserve the existing one-argument/opaque-ID contract while enforcing validation at cache/browser/request/header sinks, or make the stricter configuration contract explicit and use a release level appropriate for the breaking change. If a config is required, use a focused origin-policy type instead of requiring the unrelated `timeoutSeconds` field through `BrokerRequestConfig`, and add type/runtime compatibility tests.

- **[P2] The repository's advertised broker examples still expose the same unsafe session fields.** The core-backed TypeScript example reads a cached pending document and prints `login_url` before a validating core request; the standalone TypeScript and Python examples accept/cache/emit raw broker URLs and interpolate raw IDs/secrets into request paths and headers. Port the same trusted-origin, bounded-ID, header-value, and hostile-cache/response validation to `examples/mcp-broker-client/**`, or route the core-backed example through the validated core flow. Add regressions proving hostile values are never cached, printed/opened, or sent.

The exact head passes `pnpm check`, core tests (27/27), and `pnpm pack:dry-run`; the full-suite-only failure is the known base-branch macOS mixed-install fixture, not this diff. GitHub CI is `action_required` with zero jobs. Because this changes authentication/session handling, the zero-job exception does not apply and a real green required run is needed.

Release decision: **Patch release recommended** only if compatibility is preserved; otherwise the current patch changeset is insufficient.

mauricemohr88-debug:
Addressed the two remaining review items in `4cb8691527acd9e9403168b7e3f64050b94ab44b`.

- Preserved one-argument custom-origin normalization and safe opaque Unicode/separator IDs. Request paths stay encoded; the optional `BrokerOriginPolicy` no longer requires `timeoutSeconds`. Runtime and public-type compatibility cases are included.
- Applied origin, ID, secret, hostile-response, and hostile-cache validation to all three broker examples before cache/output/request/header operations. Equivalent IDN and IPv6 origins are accepted in Python. The related CLI status/auth-required cache display paths now also suppress invalid sessions.

Local verification passed in the existing isolated environment: `pnpm check`, `pnpm test`, package dry runs, TypeScript broker examples 17/17 each, and Python broker example 19/19. The CLI suite retains its one Windows-only skip on macOS. Independent hostile-field cases cover every validator separately. The patch changeset now includes `@call-e/core` and `@call-e/cli`; package versions are unchanged.

Required hosted CI is still a separate gate. Please approve the CI run for this head; a real green required run is needed before merge, as requested. No zero-job exception or human QA is claimed.

PR #138, opened 2026-09-14
fix(cli): report call status start as unknown

Arshgill01:
## What

`calle call status` now passes `callStarted: "unknown"` into the error envelope (the same tri-state `call start` already uses for ambiguity), instead of a literal `true`.

`cli-reference.md` now states that a status-lookup failure cannot know whether a call exists, and that `true` is reserved for a stable `run_id` from `run_call`.

## Why

The CLI reference defines `call_started: true` as: `run_call` returned a stable `run_id`. `call status` reused the stage helper with `callStarted: true`, so a fresh machine with no token reported that a call had been started:

```
npx -y @call-e/cli@0.5.1 call status --run-id run_does_not_exist --cache-root <empty> --no-telemetry --json
{
  "ok": false,
  "error": { "code": "auth_required", "message": "A usable CALL-E auth token is required." },
  "stage": "get_call_run",
  "call_started": true,
  "retry_safe": true
}
```

No call was started; the command never reached the server with credentials. `retry_safe: true` stays correct (read-only).

## Evidence

Unit test, fake cache, no network, no phone call.

```
pnpm --filter @call-e/cli test:unit
# tests 62
# pass 61
# skipped 1

# Subtest: call status auth failure reports call_started unknown
ok 40 - call status auth failure reports call_started unknown
```

`call_started` is `"unknown"`, not `true`.

## Test

`pnpm --filter @call-e/cli test:unit` and `pnpm --filter @call-e/cli check`.

## Release decision

Patch release recommended for `@call-e/cli`. Changeset included. Fixes documented error-envelope guidance; no new flag.

## Source

From the CALL-E hackathon feedback lab: https://github.com/Arshgill01/ExactRef/blob/main/docs/feedback/lab/XR-609-call-status-hardcodes-call-started-true.md

Made with [Cursor](https://cursor.com)


Arshgill01:
Friendly nudge — quiet since Sep 14; mergeable aside from CI. Happy to adjust if anything looks off.

Arshgill01:
@Ray-56 soft first-review ask — `call status` auth failures now report `call_started: \"unknown\"` instead of hardcoding `true`. Happy to adjust. No rush.

PR #139, opened 2026-09-14
fix(cli): add opt-in --redact-confirm-token for call plan

Arshgill01:
## What

`calle call plan` still prints `confirm_token` by default. `--redact-confirm-token` (or `CALLE_REDACT_CONFIRM_TOKEN=1`) removes it from `structuredContent` and every `content[].text` (JSON and plain string), sets `has_confirm_token: true`, and stores `plan_id → confirm_token` in the same private recovery cache `call recover` uses (`0600` file, `0700` dir, `schema_version: 1`).

`calle call run --plan-id X` without `--confirm-token` looks up that cache. A hit proceeds and reports `confirm_token_source: "private_cache"`. A miss keeps the existing `Missing required --confirm-token` error and adds one sentence pointing at `--redact-confirm-token`. Supplying `--confirm-token` is unchanged.

## Why

Every shipped skill documents `call plan` → read the token → `call run --plan-id … --confirm-token …`. Claude, for example:

```
2. Read the returned `plan_id` and `confirm_token`.
3. If the user's request is to place a call, immediately use `call run` with
   the exact `plan_id` and `confirm_token` returned by planning.
```

(`packages/claude-plugin/plugin/skills/calle/SKILL.md:125-128`)

Redacting the token from `call plan` stdout by default would break that flow for every agent following the skills. This PR does not edit skill text.

The same credential concern is real: `confirm_token` is what turns a plan into a dial. `auth status` already hides the bearer token (prints `cache_path` / `usable` / `expires_at`, never the access token). Recovery already keeps `plan_id` + `confirm_token` in a `0600` file and does not print them. This flag is the opt-in path that matches those two patterns without changing the documented plan → run handshake.

## Non-breaking by design

- Default `call plan` stdout is unchanged (token still printed; no `has_confirm_token`; no cache write).
- `call run --confirm-token` is unchanged.
- Skills keep working without any update.
- Flip-the-default (redact unless asked) is a later **minor**, after the skills are updated to the cache path. This PR stays **patch**.

## Tests

`pnpm --filter @call-e/cli test` and `pnpm run check`.

```
# unit
# tests 66
# pass 65
# skipped 1

# e2e
# tests 28
# pass 28

# check
Branch name follows docs/git-naming-conventions.md: fix/redact-plan-confirm-token
```

Coverage added: default still prints; `--redact-confirm-token` redacts JSON and plain-text content and writes the cache; `CALLE_REDACT_CONFIRM_TOKEN=1`; `call run --plan-id` reads the cache and reports `confirm_token_source: "private_cache"`; missing cache keeps `Missing required --confirm-token` and mentions `--redact-confirm-token`. No live `run_call`. Fixture token is `confirm-SECRET-1`.

## Source

From the CALL-E hackathon feedback lab: https://github.com/Arshgill01/ExactRef/blob/main/docs/feedback/lab/XR-604-confirm-token-printed-to-stdout.md

Made with [Cursor](https://cursor.com)


Arshgill01:
Friendly nudge — quiet since Sep 14; mergeable aside from CI. Happy to adjust if anything looks off.

Arshgill01:
@Ray-56 soft first-review ask — opt-in `--redact-confirm-token` for `call plan` (default unchanged). Happy to adjust. No rush.

